PR #2, opened 2016-09-24
Fix some wrong keys in mappings

meew0:
No huge problems, but it would cause annoying messages to be printed to console every time.

### DIFF
--- a/src/discordcr/client.cr
+++ b/src/discordcr/client.cr
@@ -113,10 +113,13 @@ module Discord
             puts "Unsupported message: #{message}"
           end
         rescue ex : JSON::ParseException
-          puts "An exception occurred during message parsing!"
-          puts ex.message
+          puts "An exception occurred during message parsing! Please report this."
+          ex.inspect_with_backtrace(STDOUT)
           puts "Raised with packet:"
           puts message
+        rescue ex
+          puts "A miscellaneous exception occurred during message handling."
+          ex.inspect_with_backtrace(STDOUT)
         end
 
         # Set the sequence to confirm that we have handled this packet, in case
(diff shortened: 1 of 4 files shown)


src/discordcr/client.cr
```diff
@@ -94,18 +94,25 @@ module Discord
...
+            handle_dispatch(packet.event_type, packet.data)
+          when OP_RECONNECT
+            handle_reconnect
+          when OP_INVALID_SESSION
+            handle_invalid_session
+          else
+            puts "Unsupported message: #{message}"
+          end
+        rescue ex
+          puts "An exception occurred during message parsing!"
```
RX14: Is this really just during message parsing? If a handler raises *any* exception won't it filter up to here? This includes user handlers and caching exceptions.

meew0: You're right, I should change it so it only triggers on JSON parse exceptions.

src/discordcr/client.cr
```diff
@@ -109,10 +109,13 @@ module Discord
             puts "Unsupported message: #{message}"
           end
         rescue ex : JSON::ParseException
-          puts "An exception occurred during message parsing!"
+          puts "An exception occurred during message parsing! Please report this."
           puts ex.message
```
RX14: Use `inspect_with_backtrace` here too.

src/discordcr/client.cr
```diff
@@ -238,7 +241,14 @@ module Discord
 
     # :nodoc:
     macro call_event(name, payload)
-      @on_{{name}}_handlers.try &.each { |handler| handler.call({{payload}}) }
+      @on_{{name}}_handlers.try &.each do |handler|
+        begin
+          handler.call({{payload}})
+        rescue ex
+          puts "An error occurred in an event handler!"
```
RX14: "An exception occurred in a user-defined event handler!"

PR #17, opened 2016-09-25
Use Travis more

meew0:
This should:

- check spec on push to any branch and PR
- check code formatting on push to any branch and PR
- build docs on push to master


deploy.sh
```diff
@@ -0,0 +1,71 @@
...
+mkdir -p out/$SOURCE_BRANCH
+
+# Clean out existing contents
+rm -rf out/$SOURCE_BRANCH/**/* || exit 0
+
+# Run our compile script
+doCompile
+
+# Move results
+mv doc/* out/$SOURCE_BRANCH/
```
RX14: Don't we want to place it in `out/doc/master`.

Also we probably want to check if this is a tag build, and if so build the tag docs.

RX14:
Also we want to delete the old doc script if we're doing this.

RX14:
Lets see if it works :P

PR #21, opened 2016-10-07
Provide and use a logger

meew0:


azah:
I wrote [a thing](https://github.com/azah/crystal.lager) for logging. Imo the default formatter is fairly obtuse, and it doesn't write to a file afaik. File logging in particular would be nice for discordcr.

meew0:
You can use an arbitrary output stream for the default logger, so not writing to file by default is not a problem. I won't use this by default but users can always overwrite it.

azah:
Fair enough. For an arbitrary output stream, how would you construct an IO that writes to both the console & a file then?

PR #23, opened 2016-10-12
Added an example called multicommand.cr

ColinRioux:
Multicommand.cr features:
» 4 commands, each with different uses. 
» Uses a command dispatcher
» Is annotated.


neko:
use two spaces for crystal indentation and add a newline to the end of the file

RX14:
Or just run the formatter

ColinRioux:
Okay, I seem to still be having formatting issues, but will get on it. (Thought it was formatting related)

RX14:
I would love to merge a more complex example, but I think it's a little too similar to the existing one at the moment. If you add a simple command dispatch using a case statement I think that would be an interesting example. Maybe showcase a few more basic things you might want to do in a bot.

PR #25, opened 2016-10-12
Added welcome bot example

neko:


ColinRioux:
This does make use of a different event, and shows how to get the channel_id from a payload of a different event, which is helpful. 

Not sure of meew0's and RX14's standards, but I think it works.

meew0:
This should cache the guild using `Cache` instead of making the request every time, to avoid API spam.

neko:
@meew0 done

PR #28, opened 2016-10-16
Simply cache initialisation

RX14:
The cache getter on Client now returns Cache and raises on nil. Use cache?
to get it if you're unsure if it's present.

Parameters other than token and client_id in the client constructor now
need to be named.


RX14:
Closed in favour of #31.

PR #29, opened 2016-10-18
Fix the format for converting discords Timestamps to Time

greenbigfrog:
(I'm just assuming that discord uses the same formatting for time in the Gateway)

### DIFF
--- a/src/discordcr/mappings/channel.cr
+++ b/src/discordcr/mappings/channel.cr
@@ -8,7 +8,7 @@ module Discord
       id: {type: UInt64, converter: SnowflakeConverter},
       channel_id: {type: UInt64, converter: SnowflakeConverter},
       author: User,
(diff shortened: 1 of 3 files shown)


src/discordcr/mappings/channel.cr
```diff
@@ -8,7 +8,7 @@ module Discord
       id: {type: UInt64, converter: SnowflakeConverter},
       channel_id: {type: UInt64, converter: SnowflakeConverter},
       author: User,
-      timestamp: {type: Time, converter: Time::Format::ISO_8601_DATE},
+      timestamp: {type: Time, converter: Time::Format.new("%FT%T.%L%:z")},
```
RX14: This should be extracted into a constant so that it isn't instantiated on every parse.

PR #30, opened 2016-10-18
add ping with response time example

greenbigfrog:


examples/ping_with_respond_time.cr
```diff
@@ -0,0 +1,18 @@
...
+
+# Make sure to replace this fake data with actual data when running.
+client = Discord::Client.new(token: "Bot MjI5NDU5NjgxOTU1NjUyMzM3.Cpnz31.GQ7K9xwZtvC40y8MPY3eTqjEIXm", client_id: 229459681955652337_u64)
+
+client.on_message_create do |payload|
+  if payload.content.starts_with? "!ping"
+    # We first create a new Message, and then we check how long it took to send the message by comparing it to the current time
+    m = client.create_message(payload.channel_id, "Pong!")
+    time = Time.utc_now - payload.timestamp
+    client.edit_message(m.channel_id, m.id, "Pong! Time taken: #{time.seconds}.#{time.milliseconds} seconds.")
```
RX14: Use a time format here

neko:
imo this could go inside the multicommand example, as it's not a very significant example

ColinRioux:
@neko Or it could go with the current ping example (`ping.cr`), since they are both ping examples. Unless `ping.cr` is meant to be simplistic and easy, then this example would fit better in `multicommand.cr`.

RX14:
Ping example is meant to be the first, simple, example: listen for x, respond with y.

examples/ping_with_respond_time.cr
greenbigfrog: Not possible for TIme Spans (I'll push other way/fix in a minute)

PR #37, opened 2016-10-20
Implemented Generic Cache

ColinRioux:
Meant to address #32 . Not sure if this was the idea, but here was my interpretation. @RX14

### DIFF
--- a/src/discordcr/cache.cr
+++ b/src/discordcr/cache.cr
@@ -34,8 +34,9 @@ module Discord
   # refrain from caching at all.
   abstract class GenericCache
     property :client, :users, :channels,
-             :guilds, :members, :roles,
-             :dm_channels, :guild_roles, :guild_channels
+      :guilds, :members, :roles,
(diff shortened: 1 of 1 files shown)


RX14:
I would merge this (albeit with a few tweaks) if not for #31. I want to do this and #31 at the same time, sorry.

meew0:
Why would `Cache` inherit from `GenericCache`, though? What I meant is the `Cache` class is simply renamed to `GenericCache` and acts as, well, a generic cache that caches most of the commonly needed Discord data. In the future there may be other, more specialised, types of caches that share nothing with `GenericCache` except for the name and the fact that they listen to client events.

ColinRioux:
@meew0 A simple answer is that I misunderstood, and it's not being merged, as it also  doesn't comply with #31, and wasn't exactly what you both were looking for. Will work on it, and will send you both my fork instead of making a pull request next time so you both can look at it.

PR #40, opened 2016-10-23
Fix the typing of GamePlaying#type

### LINKED ISSUE #27: The GamePlaying mapping doesn't consider string types
Apparently, clients can set the type field to whatever they want, even strings. This causes problems for discordcr:

error

[Relevant code section](https://github.com/meew0/discordcr/blob/master/src/discordcr/mappings/guild.cr#L134)

meew0:
Fixes #27.

### DIFF
--- a/src/discordcr/mappings/guild.cr
+++ b/src/discordcr/mappings/guild.cr
@@ -131,7 +131,7 @@ module Discord
 
     JSON.mapping(
       name: {type: String, nilable: true},
-      type: {type: UInt8, nilable: true},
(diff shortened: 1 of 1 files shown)


RX14:
The docs state:
>null, or an object containing one key of "name"  
Does it make more sense to just use a different mapping instead of altering this one, or is it just the docs being out of date.

meew0:
It's the docs being out of date; with streaming statuses `type` and `url` are also valid keys that should be parsed.

PR #42, opened 2016-11-06
Add some aliases for alternate spellings

meew0:


RX14:
We could have a compile-time flag to turn these off :P

RX14:
He actually did it the absolute madman.

RX14:
Ok, I think we need to think about our stance on this issue more carefully. There are many different methods on the raw objects that we could "crystalize", for example adding `?` to query-ish methods. Should we also explore adding simple query methods such as this?
```cr
def guild?
  !@guild_id.nil?
end
```

The question is where we draw the line between being helpful and being low-level. All of these things I've suggested add no runtime cost.

meew0:
I agree that this should be considered, however it seems out of the scope of this PR. This PR is just intended to decrease confusion, these other suggestions would be for utility.

RX14:
Yeah, let's just squash and merge this.

PR #46, opened 2016-12-31
Reactions

meew0:
Includes:

- REST methods to add reactions
- Events for added reactions


RX14:
do we even have regular emojii support? IDEK how that works

meew0:
We have regular emoji support but the reaction endpoints use a special string-only format which is documented in the method doc comments. We might want a helper method to turn a regular `Emoji`/`ReactionEmoji` object into the special format but it's not really that important.

PR #52, opened 2017-03-01
Add a REST method for uploading files

meew0:
This wasn't possible to do before due to a lack of HTTP form data support in the stdlib.

### DIFF
--- a/src/discordcr/rest.cr
+++ b/src/discordcr/rest.cr
@@ -343,7 +343,7 @@ module Discord
     def upload_file(channel_id : UInt64, content : String?, file : IO, filename : String? = nil)
       io = IO::Memory.new
 
-      if filename.nil?
+      unless filename
         if file.is_a? File
           filename = File.basename(file.path)
         else
--- a/src/discordcr/rest.cr
+++ b/src/discordcr/rest.cr
@@ -1,4 +1,5 @@
 require "http/client"
+require "http/formdata"
 require "openssl/ssl/context"
 require "time/format"
 
@@ -335,20 +336,30 @@ module Discord
     # [API docs for this method](https://discordapp.com/developers/docs/resources/channel#create-message)
     # (same as `#create_message` -- this method implements form data bodies
     # while `#create_message` implements JSON bodies)
-    def upload_file(channel_id : UInt64, content : String?, file : File)
+    def upload_file(channel_id : UInt64, content : String?, file : IO, filename : String? = nil)
       io = IO::Memory.new
-      boundary = "--------------------------#{SecureRandom.urlsafe_base64(18)}"
-      HTTP::FormData.generate(io, boundary) do |g|
-        g.field("content", content) if content
-        g.file("file", file, HTTP::FormData::FileMetadata.new(filename: File.basename(file.path)))
+
+      if filename.nil?
+        if file.is_a? File
+          filename = File.basename(file.path)
+        else
+          filename = ""
+        end
       end
+
+      builder = HTTP::FormData::Builder.new(io)
+
+      builder.field("content", content) if content
+      builder.file("file", file, HTTP::FormData::FileMetadata.new(filename: filename))
+
+      builder.finish
 
       response = request(
         :channels_cid_messages,
         channel_id,
         "POST",
         "/channels/#{channel_id}/messages",
-        HTTP::Headers{"Content-Type" => "multipart/form-data; boundary=#{boundary}"},
+        HTTP::Headers{"Content-Type" => builder.content_type},
         io.to_s
       )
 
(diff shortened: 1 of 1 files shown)


src/discordcr/rest.cr
```diff
@@ -329,7 +329,33 @@ module Discord
...
-    # TODO: Add the upload file endpoint when the multipart PR is merged
+    # Uploads a file to a channel. Requires the "Send Messages" and "Attach
+    # Files" permissions.
+    #
+    # [API docs for this method](https://discordapp.com/developers/docs/resources/channel#create-message)
+    # (same as `#create_message` -- this method implements form data bodies
+    # while `#create_message` implements JSON bodies)
+    def upload_file(channel_id : UInt64, content : String?, file : File)
+      io = IO::Memory.new
+      boundary = "--------------------------#{SecureRandom.urlsafe_base64(18)}"
```
RX14: Unneccesary, this is the same as the default.

src/discordcr/rest.cr
```diff
@@ -329,7 +329,33 @@ module Discord
...
+      HTTP::FormData.generate(io, boundary) do |g|
```
RX14: It's been `.build` since it got merged into the stdlib.

src/discordcr/rest.cr
```diff
@@ -329,7 +329,33 @@ module Discord
...
+        g.field("content", content) if content
+        g.file("file", file, HTTP::FormData::FileMetadata.new(filename: File.basename(file.path)))
```
RX14: Please allow more flexibility: people should be able to pass an `IO` instead of a `File`, and use custom filenames.

src/discordcr/rest.cr
```diff
@@ -335,20 +336,30 @@ module Discord
...
-      boundary = "--------------------------#{SecureRandom.urlsafe_base64(18)}"
-      HTTP::FormData.generate(io, boundary) do |g|
-        g.field("content", content) if content
-        g.file("file", file, HTTP::FormData::FileMetadata.new(filename: File.basename(file.path)))
+
+      if filename.nil?
+        if file.is_a? File
+          filename = File.basename(file.path)
+        else
+          filename = ""
```
RX14: This should be `nil` so that the `filename` simply isn't attached.

src/discordcr/rest.cr
```diff
@@ -335,20 +336,30 @@ module Discord
...
     # while `#create_message` implements JSON bodies)
-    def upload_file(channel_id : UInt64, content : String?, file : File)
+    def upload_file(channel_id : UInt64, content : String?, file : IO, filename : String? = nil)
       io = IO::Memory.new
-      boundary = "--------------------------#{SecureRandom.urlsafe_base64(18)}"
-      HTTP::FormData.generate(io, boundary) do |g|
-        g.field("content", content) if content
-        g.file("file", file, HTTP::FormData::FileMetadata.new(filename: File.basename(file.path)))
+
+      if filename.nil?
```
RX14: Document this logic in the method description, it's lib-specific.

RX14: With those 2 suggestions combined, the code simply becomes:
```cr
filename = File.basename(file.path) if filename.nil? && file.is_a? File
```

src/discordcr/rest.cr
```diff
@@ -335,20 +336,30 @@ module Discord
...
+        if file.is_a? File
+          filename = File.basename(file.path)
+        else
+          filename = ""
+        end
       end
 
+      builder = HTTP::FormData::Builder.new(io)
+
```
RX14: This whitespace can go.

src/discordcr/rest.cr
meew0: If it is `nil` the sending will not work at all as Discord expects the property to be sent; if it's an empty string a placeholder filename will be generated.

RX14: Fucking trashware.

src/discordcr/rest.cr
```diff
@@ -329,7 +330,45 @@ module Discord
...
+    # the file's filename will be used instead. If it is an `IO` without
+    # filename information, Discord will generate a placeholder filename.
+    #
+    # [API docs for this method](https://discordapp.com/developers/docs/resources/channel#create-message)
+    # (same as `#create_message` -- this method implements form data bodies
+    # while `#create_message` implements JSON bodies)
+    def upload_file(channel_id : UInt64, content : String?, file : IO, filename : String? = nil)
+      io = IO::Memory.new
+
+      if filename.nil?
```
RX14: In that case, use `unless filename`

RX14:
Do I bully @meew0 too much in code reviews? Cast your votes below.

PR #56, opened 2017-04-19
Basic voice sending

meew0:
This is a fully working implementation of voice sending to Discord. I have tested it and it worked without problems, however it would be useful if others would test this on their systems, given that it introduces a native dependency.

So far, only sending already-encoded Opus frames is supported, ~~and there is no automatic timing – the timing has to be done manually.~~ **(Automatic timing has been implemented)**

It should be used approximately like this **(Out of date, see the included example for a working demo)**:

```cr
# somewhere
client.voice_state_update(guild_id, channel_id, false, false)

# A VOICE_SERVER_UPDATE dispatch is received as a response to the sent voice_state_update
client.on_voice_server_update do |payload|
  # The payload can be used to create a voice client
  voice_client = Discord::VoiceClient.new(payload, client.session, user_id)

  # Called on successful connection
  voice_client.on_ready do
    # Tell Discord that we are speaking (green circle in user list). This is
    # important and has to be done at least once in every connection, because
    # Discord uses this packet to tell its clients which SSRC belongs to whom.
    voice_client.send_speaking(true)

    loop do
      opus_data = magically_produce_an_opus_frame()
      voice_client.play_opus(opus_data)

      # This timing is slightly inaccurate; ideally the time it takes to send
      # the packet would be measured and then we would sleep 20 ms - that time.
      sleep 20.milliseconds
    end
  end
end
```

Questions:
- How many more abstractions should we provide? ~~Accurate timing, opus encoding, DCA reading, bindings to ffmpeg for playing arbitrary files? I'm thinking we should definitely do the first one, but the others probably don't fit the concept of the lib.~~ **We've decided on accurate timing and DCA reading**
- Should the `play_opus` method keep its name? ~~It's not the ideal name in my opinion but I 
can't really think of an alternative.~~ **Yes**
- Should `discordcr.cr` be changed to not automatically require `voice.cr`, so the user doesn't have to incur the libsodium dependency if they don't want voice? ~~I would say yes but I'm not sure how acceptable that is to do in Crystal.~~ **Crystal does this automatically if the user doesn't use voice code**

### DIFF
--- a/examples/voice_send.cr
+++ b/examples/voice_send.cr
@@ -1,0 +1,130 @@
+# This is a simple music bot that can connect to a voice channel and play back
+# some music in DCA format. It demonstrates how to use VoiceClient and
+# DCAParser.
+#
+# For more information on the DCA file format, see
+# https://github.com/bwmarrin/dca.
+
+require "../src/discordcr"
+
+# Make sure to replace this fake data with actual data when running.
+client = Discord::Client.new(token: "Bot MjI5NDU5NjgxOTU1NjUyMzM3.Cpnz31.GQ7K9xwZtvC40y8MPY3eTqjEIXm", client_id: 229459681955652337_u64)
+
+# ID of the current user, required to create a voice client
+current_user_id = nil
+
+# The ID of the (text) channel in which the connect command was run, so the
+# "Voice connected." message is sent to the correct channel
+connect_channel_id = nil
+
+# Where the created voice client will eventually be stored
+voice_client = nil
+
+client.on_ready do |payload|
+  current_user_id = payload.user.id
+end
+
+client.on_message_create do |payload|
+  if payload.content.starts_with? "!connect "
+    # Used as:
+    # !connect <guild ID> <channel ID>
+
+    # Parse the command arguments
+    ids = payload.content[9..-1].split(' ').map(&.to_u64)
+
+    client.create_message(payload.channel_id, "Connecting...")
+    connect_channel_id = payload.channel_id
+    client.voice_state_update(ids[0].to_u64, ids[1].to_u64, false, false)
+  elsif payload.content.starts_with? "!play_dca "
+    # Used as:
+    # !play_dca <filename>
+    #
+    # Make sure the DCA file you play back is valid according to the spec
+    # (including metadata), otherwise playback will fail.
+
+    unless voice_client
+      client.create_message(payload.channel_id, "Voice client is nil!")
+      next
+    end
+
+    filename = payload.content[10..-1]
+    file = File.open(filename)
+
+    # The DCAParser class handles parsing of the DCA file. It doesn't do any
+    # sending of audio data to Discord itself – that has to be done by
+    # VoiceClient.
+    parser = Discord::DCAParser.new(file)
+
+    # A proper DCA(1) file contains metadata, which is exposed by DCAParser.
+    # This metadata may be of interest, so here is some example code that uses
+    # it.
+    if metadata = parser.metadata
+      tool = metadata.dca.tool
+      client.create_message(payload.channel_id, "DCA file was created by #{tool.name}, version #{tool.version}.")
+
+      if info = metadata.info
+        client.create_message(payload.channel_id, "Song info: #{info.title} by #{info.artist}.") if info.title && info.artist
+      end
+    else
+      client.create_message(payload.channel_id, "DCA file metadata is invalid!")
+    end
+
+    # Set the bot as speaking (green circle). This is important and has to be
+    # done at least once in every voice connection, otherwise the Discord client
+    # will not know who the packets we're sending belongs to.
+    voice_client.not_nil!.send_speaking(true)
+
+    client.create_message(payload.channel_id, "Playing DCA file `#{filename}`.")
+
+    # For smooth audio streams Discord requires one packet every
+    # 20 milliseconds. The `every` method measures the time it takes to run the
+    # block and then sleeps 20 milliseconds minus that time before moving on to
+    # the next iteration, ensuring accurate timing.
+    #
+    # When simply reading from DCA, the time it takes to read, process and
+    # send the frame is small enough that `every` doesn't make much of a
+    # difference (in fact, some users report that it actually makes things
+    # worse). If the processing time is not negligibly slow because you're
+    # doing something else than DCA parsing, or because you're reading from a
+    # slow source, or for any other reason, then it is recommended to use
+    # `every`. Otherwise, simply using a loop and `sleep`ing `20.milliseconds`
+    # each time may suffice.
+    Discord.every(20.milliseconds) do
+      frame = parser.next_frame(reuse_buffer: true)
+      break unless frame
+
+      # Perform the actual sending of the frame to Discord.
+      voice_client.not_nil!.play_opus(frame)
+    end
+
+    # Alternatively, the above code can be realised as the following:
+    #
+    # parser.parse do |frame|
+    #   Discord.timed_run(20.milliseconds) do
+    #     voice_client.not_nil!.play_opus(frame)
+    #   end
+    # end
+    #
+    # (The `parse` method reads the frames consecutively and passes them to the
+    # block.)
+
+    file.close
+  end
+end
+
+# The VOICE_SERVER_UPDATE dispatch is sent by Discord once the op4 packet sent
+# by voice_state_update has been processed. It tells the client the endpoint
+# to connect to.
+client.on_voice_server_update do |payload|
+  begin
+    vc = voice_client = Discord::VoiceClient.new(payload, client.session.not_nil!, current_user_id.not_nil!)
+    vc.on_ready do
+      client.create_message(connect_channel_id.not_nil!, "Voice connected.")
+    end
+    vc.run
+  rescue e
+    e.inspect_with_backtrace(STDOUT)
+  end
+end
+
+client.run
--- a/src/discordcr/dca.cr
+++ b/src/discordcr/dca.cr
@@ -1,0 +1,150 @@
+require "json"
+
+module Discord
+  # Parser for the DCA file format, a simple wrapper around Opus made
+  # specifically for Discord bots.
+  class DCAParser
+    # Magic string that identifies a DCA1 file
+    DCA1_MAGIC = "DCA1"
+
+    # The parsed metadata, or nil if it could not be parsed.
+    getter metadata : DCA1Mappings::Metadata?
+
+    # Create a new parser. It will read from the given *io*. If *raw* is set,
+    # the file is assumed to be a DCA0 file, without any metadata. If the file's
+    # metadata doesn't conform to the DCA1 specification and *strict_metadata*
+    # is set, then the parsing will fail with an error; if it is not set then
+    # the metadata will silently be `nil`.
+    def initialize(@io : IO, raw = false, @strict_metadata = true)
+      unless raw
+        verify_magic
+        parse_metadata
+      end
+    end
+
+    # Reads the next frame from the IO. If there is nothing left to read, it
+    # will return `nil`.
+    #
+    # If *reuse_buffer* is true, a large buffer will be allocated once and
+    # reused for future calls of this method, reducing the load on the GC and
+    # potentially reusing memory use overall; if it is false, a new buffer of
+    # just the correct size will be allocated every time. Note that if the
+    # buffer is reused, the returned data is only valid until the next call to
+    # `next_frame`.
+    def next_frame(reuse_buffer = false) : Bytes?
+      begin
+        header = @io.read_bytes(Int16, IO::ByteFormat::LittleEndian)
+        raise "Negative frame header (#{header} < 0)" if header < 0
+
+        buf = if reuse_buffer
+                full_buf = @reused_buffer ||= Bytes.new(Int16::MAX)
+                full_buf[0, header]
+              else
+                Bytes.new(header)
+              end
+
+        @io.read_fully(buf)
+        buf
+      rescue IO::EOFError
+        nil
+      end
+    end
+
+    # Continually reads frames from the IO until there are none left. Each frame
+    # is passed to the given *block*.
+    def parse(&block : Bytes ->)
+      loop do
+        buf = next_frame
+
+        if buf
+          block.call(buf)
+        else
+          break
+        end
+      end
+    end
+
+    private def verify_magic
+      magic = @io.read_string(4)
+      if magic != DCA1_MAGIC
+        raise "File is not a DCA1 file (magic is #{magic}, should be DCA1)"
+      end
+    end
+
+    private def parse_metadata
+      # The header of the metadata part is the four-byte size of the following
+      # metadata payload.
+      metadata_size = @io.read_bytes(Int32, IO::ByteFormat::LittleEndian)
+      metadata_io = IO::Sized.new(@io, read_size: metadata_size)
+
+      begin
+        @metadata = DCA1Mappings::Metadata.from_json(metadata_io)
+      rescue e : JSON::ParseException
+        raise e if @strict_metadata
+      end
+
+      metadata_io.skip_to_end
+    end
+  end
+
+  # Mappings for DCA1 metadata
+  module DCA1Mappings
+    struct Metadata
+      JSON.mapping(
+        dca: DCA,
+        opus: Opus,
+        info: {type: Info, nilable: true},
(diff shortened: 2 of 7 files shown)


src/discordcr/voice.cr
```diff
@@ -0,0 +1,268 @@
...
+    OP_SELECT_PROTOCOL     = 1
+    OP_READY               = 2
+    OP_HEARTBEAT           = 3
+    OP_SESSION_DESCRIPTION = 4
+    OP_SPEAKING            = 5
+    OP_HELLO               = 8
+
+    # The heartbeat is the same every time, so there's no point in constructing
+    # it from an object
+    HEARTBEAT_JSON = %({"op":#{OP_HEARTBEAT},"d":null})
```
RX14: I'd construct it from a named tuple regardless, its just kinda good practice.

src/discordcr/voice.cr
```diff
@@ -0,0 +1,268 @@
...
+    # Creates a new voice client. The *payload* should be a payload received
+    # from Discord as part of a VOICE_SERVER_UPDATE dispatch, received after
+    # sending a voice state update (gateway op 4) packet. The *session* should
+    # be the session currently in use by the gateway client on which the
+    # aforementioned dispatch was received, and the *user_id* should be the
+    # user ID of the account on which the voice client is created. (It is
+    # received as part of the gateway READY dispatch, for example)
+    def initialize(payload : Discord::Gateway::VoiceServerUpdatePayload,
+                   session : Discord::Gateway::Session, @user_id : UInt64)
+      @endpoint = payload.endpoint.gsub(":80", "")
```
RX14: This might break with ipv6, so just in case parse this with URI.parse

src/discordcr/voice.cr
```diff
@@ -0,0 +1,268 @@
...
+    @secret_key : Array(UInt8)?
+    property secret_key
+
+    def initialize
+      @socket = UDPSocket.new
+    end
+
+    def connect(endpoint : String, port : UInt32, ssrc : UInt32)
+      endpoint = endpoint[6..-1] if endpoint.starts_with? "wss://"
+      endpoint = endpoint.gsub(":80", "") # The endpoint may contain a port, we don't want that
```
RX14: Same here, use the uri parser.

src/discordcr/voice.cr
```diff
@@ -0,0 +1,268 @@
...
+      @socket.connect(endpoint, port)
+    end
+
+    # Sends a discovery packet to Discord, telling them that we want to know our
+    # IP so we can select the protocol on the VWS
+    def send_discovery
+      io = IO::Memory.new(70)
+
+      io.write_bytes(@ssrc.not_nil!, IO::ByteFormat::BigEndian) # [@ssrc].pack("N")
+      io.write(Bytes.new(66, 0_u8))                             # 66 null bytes for a total of 70 bytes
```
RX14: Id use 70 - sizeof(Int32) here. It's a bit more self-explanitory. Also can we loose the ruby translations, its more than possible to see what this code does without looking at the docs (unlike ruby).

src/discordcr/voice.cr
```diff
@@ -0,0 +1,268 @@
...
+    # Awaits a response to the discovery request and returns our local IP and
+    # port once the response is received
+    def receive_discovery_reply : {String, UInt16}
+      buf = Bytes.new(70)
+      @socket.receive(buf)
+      io = IO::Memory.new(buf)
+
+      io.seek(4)                                              # The first four bytes are just the SSRC again, we don't care about that
+      ip = io.read_string(64).delete("\0")                    # Our IP
+      port = io.read_bytes(UInt16, IO::ByteFormat::BigEndian) # Our port
```
RX14: These last 2 comments are pointless. I can see what variable they are being assigned to.

src/discordcr/voice.cr
```diff
@@ -0,0 +1,268 @@
...
+    def send_audio(buf, sequence, time)
+      header = create_header(sequence, time)
+
+      buf = encrypt_audio(header, buf)
+
+      new_buf = Bytes.new(header.size + buf.size)
+      header.copy_to(new_buf)
+      buf.copy_to(new_buf + header.size)
+
+      @socket.write(new_buf)
```
RX14: Tfw no writev

src/discordcr/voice.cr
```diff
@@ -0,0 +1,268 @@
...
+
+      # Prepend the buf with zero_bytes zero bytes
+      message = Bytes.new(buf.size + zero_bytes, 0_u8)
+      buf.copy_to(message + zero_bytes)
+
+      # Create a buffer for the ciphertext
+      c = Bytes.new(message.size)
+
+      # Encrypt
+      key = Bytes.new(@secret_key.not_nil!.to_unsafe, @secret_key.not_nil!.size)
```
RX14: Array should have a to_slice, and why not use a slice instead of an array to hold the key anyway.

src/discordcr/websocket.cr
```diff
@@ -0,0 +1,80 @@
...
+        payload = parse_message(message)
+        handler.call(payload)
+      end
+    end
+
+    def on_close(&handler : String ->)
+      @websocket.on_close(&handler)
+    end
+
+    def run
```
RX14: Use the delegate macro for these.

RX14:
Pls use fix up commits

src/discordcr/voice.cr
meew0: ?

src/discordcr/voice.cr
meew0: Well, they don't. Discord sends it as an array so it will fundamentally be an array somewhere, but you're right that it's more efficient to do the to_unsafe thing once rather than for every packet

src/discordcr/voice.cr
meew0: I can't use the URI parser because the endpoint doesn't have a scheme (it's usually just `eu-west1.discord.gg:80` or something)

src/discordcr/voice.cr
meew0: Same as above

src/discordcr/dca.cr
```diff
@@ -0,0 +1,135 @@
...
+
+    # The parsed metadata, or nil if it could not be parsed.
+    getter metadata : DCA1Mappings::Metadata?
+
+    # Create a new parser. It will read from the given *io*. If *raw* is set,
+    # the file is assumed to be a DCA0 file, without any metadata. If the file's
+    # metadata doesn't conform to the DCA1 specification and *strict_metadata*
+    # is set, then the parsing will fail with an error; if it is not set then
+    # the metadata will silently be `nil`.
+    def initialize(@io : IO, raw = false, @strict_metadata = false)
```
RX14: Strictness should be on by default.

src/discordcr/dca.cr
```diff
@@ -0,0 +1,135 @@
...
+
+    private def verify_magic
+      magic = @io.read_string(4)
+      if magic != DCA1_MAGIC
+        raise "File is not a DCA1 file (magic is #{magic}, should be DCA1)"
+      end
+    end
+
+    private def parse_metadata
+      header = @io.read_bytes(Int32, IO::ByteFormat::LittleEndian)
```
RX14: `header_size`

src/discordcr/dca.cr
```diff
@@ -0,0 +1,135 @@
...
+      sized = IO::Sized.new(@io, read_size: header)
```
RX14: `header_io`

src/discordcr/dca.cr
```diff
@@ -0,0 +1,135 @@
...
+    end
+
+    # Reads the next frame from the IO. If there is nothing left to read, it
+    # will return `nil`.
+    def next_frame : Bytes?
+      begin
+        header = @io.read_bytes(Int16, IO::ByteFormat::LittleEndian)
+        raise "Negative frame header (#{header} < 0)" if header < 0
+
+        buf = Bytes.new(header)
```
RX14: The size of this buffer is restricted by `Int16::Max` (and it's small), so it would be pretty sensible I think to allocate this buffer once and reuse it. However this would mean the slice returned by this method would be only valid until the next call. Perhaps an option `reuse_buffer = false` for this method and lazy allocate the buffer?

examples/voice_send.cr
```diff
@@ -0,0 +1,119 @@
...
+  end
+end
+
+# The VOICE_SERVER_UPDATE dispatch is sent by Discord once the op4 packet sent
+# by voice_state_update has been processed. It tells the client the endpoint
+# to connect to.
+client.on_voice_server_update do |payload|
+  begin
+    voice_client = Discord::VoiceClient.new(payload, client.session.not_nil!, current_user_id.not_nil!)
+    voice_client.not_nil!.on_ready do
```
RX14: use a dual local/global assignment to avoid these `not_nil!` calls. For example `vc = voice_client = VC.new(...)`

examples/voice_send.cr
```diff
@@ -0,0 +1,119 @@
...
+      #
+      # When simply reading from DCA, the time it takes to read, process and
+      # send the frame is small enough that `timed_run` doesn't make much of a
+      # difference (in fact, some users report that it actually makes things
+      # worse). If the processing time is not negligibly slow because you're
+      # doing something else than DCA parsing, or because you're reading from a
+      # slow source, or for any other reason, then it is recommended to use
+      # `timed_run`. Otherwise, simply doing something like
+      # `sleep 20.milliseconds` may suffice.
+      Discord::VoiceClient.timed_run do
```
RX14: This should use `Discord.every`.

examples/voice_send.cr
```diff
@@ -0,0 +1,119 @@
...
+
+    # The DCAParser class handles parsing of the DCA file. It doesn't do any
+    # sending of audio data to Discord itself – that has to be done by
+    # VoiceClient.
+    parser = Discord::DCAParser.new(file)
+
+    # A proper DCA(1) file contains metadata, which is exposed by DCAParser.
+    # This metadata may be of interest, so here is some example code that uses
+    # it.
+    if parser.metadata
```
RX14: `if metadata = parser.metadata` dear god.

examples/voice_send.cr
```diff
@@ -0,0 +1,119 @@
...
+      metadata = parser.metadata.not_nil!
+      tool = metadata.dca.tool
+      client.create_message(payload.channel_id, "DCA file was created by #{tool.name}, version #{tool.version}.")
+
+      if metadata.info
```
RX14: ditto

src/discordcr/voice.cr
```diff
@@ -90,6 +90,19 @@ module Discord
       @udp.send_audio(buf, @sequence, @time)
     end
 
+    # Utility function that runs the given block and measures the time it takes,
+    # then sleeps the given time (or 20 ms by default) minus that time. This is
+    # useful because (in most cases) voice data should be sent to Discord at a
+    # rate of one frame every 20 ms, and if the processing and sending takes
+    # a certain amount of time, then noticeable choppiness can be heard.
+    def self.timed_run(total_time = 20.milliseconds, &block)
```
RX14: Delet this

src/discordcr/dca.cr
meew0: No; according to the [DCA1 spec](https://github.com/bwmarrin/dca/wiki/DCA1-specification#header) the size *is* the header of the metadata section.

src/discordcr/dca.cr
meew0: Same here, except [link to different part of the spec](https://github.com/bwmarrin/dca/wiki/DCA1-specification#header-1)

src/discordcr/dca.cr
RX14: Call it `size` then as the method gives context. `header` is ambiguous.

examples/voice_send.cr
meew0: But it's more compact to do it with the loop provided by `parse`, which handles nil buffers on EOF.

src/discordcr/voice.cr
meew0: See my above comment on `timed_run`

examples/voice_send.cr
RX14: No its not. A 1-line guard statement is cleaner than more indent.

src/discordcr/dca.cr
meew0: What's the advantage of reusing the buffer? I'd assume the performance impact is negligible compared to the IO read.

RX14: Less garbage means the program will use less memory and there will be less GC work done.

examples/voice_send.cr
meew0: I added both variants to the example now, so the user can decide by themselves which they want to use.

src/discordcr/voice.cr
meew0: See above again; we still need `timed_run` in case the user wants to only send one audio packet. It may not be useful in the general situation but the spirit of the lib is to give the user a lot of control, so I'd keep this method regardless.

examples/voice_send.cr
```diff
@@ -50,7 +50,7 @@ client.on_message_create do |payload|
     # The DCAParser class handles parsing of the DCA file. It doesn't do any
     # sending of audio data to Discord itself – that has to be done by
     # VoiceClient.
-    parser = Discord::DCAParser.new(file)
+    parser = Discord::DCAParser.new(file, strict_metadata: false)
```
RX14: Why?

src/discordcr/voice.cr
```diff
@@ -137,11 +137,20 @@ module Discord
     end
 
     private def on_close(message : String)
-      LOGGER.warn("VWS closed with message: #{message}")
+      unless message.bytesize >= 2
```
RX14: `if message.bytesize < 2`?

examples/voice_send.cr
```diff
@@ -97,6 +95,15 @@ client.on_message_create do |payload|
...
     end
 
+    # The above code can also equivalently be realised as:
+
+    # Discord.every(20.milliseconds) do
+    #   frame = parser.next_frame(reuse_buffer: true)
+    #   break unless frame
+    #
+    #   voice_client.not_nil!.play_opus(frame)
+    # end
```
RX14: Look how much cleaner that is. Remove `timed_run` and change the example. Even if you leave `timed_run`, `every` has a much *much* cleaner API and should be used as the default in the examples.

examples/voice_send.cr
meew0: So the example can still play DCA files if their metadata is invalid (since the metadata is irrelevant for the audio playback)

examples/voice_send.cr
meew0: In my opinion, the other version looks cleaner because it doesn't have the nil check and because the internals of how the frame is obtained are abstracted more. Why do you think this version is cleaner specifically?

examples/voice_send.cr
RX14: Why the fuck should we support files with invalid metadata anyway?

examples/voice_send.cr
RX14: Less indentation, `every` is an understandable method name compared to `timed_run`, logic is clearer.

examples/voice_send.cr
meew0: Because people might not care about it, and thus not bother to fill the data fields out as required.

RX14: Whats the point of a spec if all the implementations allow you to ignore it?

meew0: This example isn't meant to make some statement about the point of a spec, it's meant to be usable in a real world situation where people might not care about the metadata and leave it empty.

examples/voice_send.cr
```diff
@@ -73,36 +76,37 @@ client.on_message_create do |payload|
...
-    # The above code can also equivalently be realised as:
-
-    # Discord.every(20.milliseconds) do
-    #   frame = parser.next_frame(reuse_buffer: true)
-    #   break unless frame
+    # Alternatively, the above code can be realised as the following:
     #
-    #   voice_client.not_nil!.play_opus(frame)
+    # parser.parse do |frame|
+    #   Discord::VoiceClient.timed_run do
```
RX14: Please move `timed_run` to `Discord` and make it take the same argument the same way as `every`. You should use `timed_run` in the definition of `every` tbh.

RX14:
This looks good, please rebase this and i'll review the entire thing again.

PR #63, opened 2017-05-12
add Reaction, Message#reactions

z64:
`reactions` was missing from the `Message` mapping.

This adds a data object for a `Reaction`, and makes `Message` parse any reactions on a message payload into an `Array(Reaction)`.


RX14:
Should be 1 commit.

PR #64, opened 2017-05-15
Message#parse_mentions

### LINKED ISSUE #35: Stuff for parsing mentions
proposed interface:

``` cr
parse_mentions(str) do |content|
  case content
  when Discordcr::Mention
    # Mention
  when String
    # message hunk
  end
end
```

Might want a few fancier wrappers around it if required.

z64:
I went as best I could off of the talk / psuedocode in #35.

I didn't think it was really fitting to put it as a `self` method of `Discord` (`Discord.parse_mentions`) as `Message#content` is the only place they will ever occur.

I'm also not sure if [this is still desirable](https://github.com/meew0/discordcr/issues/35#issuecomment-254963274):
> [..] because you don't get their position in the string. 

I personally don't need it (did you mean the literal string index? and can't really think of a use case not covered by this PR, some kind of exotic parsing I guess?), but I can look into implementing it.

Also, not 100% sure how I feel about the overload `def parse_mentions(typ : MentionType)`.

Example usage: **(out of date, see diff)**
```crystal
msg = client.get_channel_message(246283902652645376u64, 313536439444832257u64)

msg.parse_mentions do |kind, id, name|
  case kind
  when Discord::MentionType::User
    puts "User mentioned: #{id}"
  when Discord::MentionType::Channel
    puts "Channel mentioned: #{id}"
  when Discord::MentionType::Role
    puts "Role mentioned: #{id}"
  when Discord::MentionType::Emoji
    puts "Emoji mentioned: #{name}:#{id}"
  when Discord::MentionType::Everyone
    puts "Everyone mentioned"
  when Discord::MentionType::Here
    puts "Here mentioned"
  end
end
# => User mentioned: 120571255635181568
# => User mentioned: 225375446223683606
# => Role mentioned: 244951251865960449
# => Channel mentioned: 225375815087554563
# => Emoji mentioned: old_man:246811049834053632
# => Everyone mentioned
# => Here mentioned

p msg.parse_mentions
# #=> {User => [{User, 120571255635181568, nil}, {User, 225375446223683606, nil}],
#       Role => [{Role, 244951251865960449, nil}],
#       Channel => [{Channel, 225375815087554563, nil}],
#       Emoji => [{Emoji, 246811049834053632, "old_man"}],
#       Everyone => [{Everyone, nil, nil}],
#       Here => [{Here, nil, nil}]}

p msg.parse_mentions(Discord::MentionType::User)
# => [{User, 120571255635181568, nil}, {User, 225375446223683606, nil}]
```

Closes #35

### DIFF
--- a/src/discordcr/mappings/channel.cr
+++ b/src/discordcr/mappings/channel.cr
@@ -18,10 +18,6 @@ module Discord
       pinned: {type: Bool, nilable: true}
     )
 
-    # A mention contained within a message, constructed by its type,
-    # the snowflake, and it's name if it is an emoji.
-    alias Mention = Tuple(MentionType, UInt64?, String?)
-
     # A hash map of regex describing how mentions are parsed by type
     MENTION_REGEX = {
       MentionType::User     => /<@!?(?<id>\d+)>/,
@@ -32,20 +28,12 @@ module Discord
       MentionType::Here     => /@here/,
     }
 
-    # Returns a map of all mentions contained in the message.
+    # Returns an array of all mentions contained in the message.
     def parse_mentions
-      mentions = {} of MentionType => Array(Mention)
+      mentions = [] of Mention
 
-      {% for typ in {
-                      MentionType::User,
-                      MentionType::Role,
-                      MentionType::Channel,
-                      MentionType::Emoji,
-                      MentionType::Everyone,
-                      MentionType::Here,
-                    } %}
-        mentions[MentionType.new({{typ}})] = parse_mentions(MentionType.new({{typ}}))
-      {% end %}
+      parse_mentions { |m| mentions << m }
+
       mentions
     end
 
@@ -65,6 +53,10 @@ module Discord
       end
     end
   end
+
+  # A mention contained within a message, constructed by its type,
+  # the snowflake, and its name if it is an emoji.
+  record Mention, type : MentionType, id : UInt64?, name : String?
 
   # An enum of the different kinds of Discord mentions
   enum MentionType
--- a/src/discordcr/mappings/channel.cr
+++ b/src/discordcr/mappings/channel.cr
@@ -17,6 +17,63 @@ module Discord
       embeds: Array(Embed),
       pinned: {type: Bool, nilable: true}
     )
+
+    # A mention contained within a message, constructed by its type,
+    # the snowflake, and it's name if it is an emoji.
+    alias Mention = Tuple(MentionType, UInt64?, String?)
+
+    # A hash map of regex describing how mentions are parsed by type
+    MENTION_REGEX = {
+      MentionType::User     => /<@!?(?<id>\d+)>/,
+      MentionType::Role     => /<@&(?<id>\d+)>/,
+      MentionType::Channel  => /<#(?<id>\d+)>/,
+      MentionType::Emoji    => /<:(?<name>\w+):(?<id>\d+)>/,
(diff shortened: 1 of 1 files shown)


src/discordcr/mappings/channel.cr
```diff
@@ -17,6 +17,63 @@ module Discord
       embeds: Array(Embed),
       pinned: {type: Bool, nilable: true}
     )
+
+    # A mentioned contained within a message, constructed by its type,
```
z64: Woops typo.

src/discordcr/mappings/channel.cr
```diff
@@ -17,6 +17,63 @@ module Discord
       embeds: Array(Embed),
       pinned: {type: Bool, nilable: true}
     )
+
+    # A mention contained within a message, constructed by its type,
+    # the snowflake, and it's name if it is an emoji.
+    alias Mention = Tuple(MentionType, UInt64?, String?)
```
z64: Looking at this now, I think it would be better if I described a `Mention` as a `record` struct instead.

RX14: I'd return just an array of all mention types too. It's easy enough to use `mentions.select(&.type == User)`, and for typical messages (less than 5 mentions) it'll be faster too. Worse case is only `O(n)` as well.

meew0:
Can you add specs for this? (I'm pretty late aren't I)

z64:
Yeah, I can do that. I wanted to make sure this interface was acceptable though.

meew0:
Are you still working on this?

z64:
Per my last comment, I still wasn't sure if this met the requirements you and RX had in mind in the original issue #35 . Throw me a 👍 and I'll get those specs in.

meew0:
It's fine by me, @RX14?

RX14:
Why not a `UserMention`, etc. set of structs which have exactly the available fields. In the block we just yield the object of the correct type to the block (only 1 arg) and switch on the type instead of using an enum.

`parse_mentions`  should just return an array, and we should have `user_mentions` etc. methods instead of `parse_mentions(User)`.

And this still doesn't give me the location, and size of the mention in the message text.

If I want a bot to replace `Foo Bar @mention Baz` with `Foo Bar @other_mention Baz` it has to do some manual parsing, instead of being able to get the location of the mention in the string to be able to modify and transform the string. Character index and length of each mention should be in the objects.

z64:
@RX14 Sorry, had you looked at the code? The example in my OP is out of date. `#parse_mentions` does return an array, and only one block arg is yielded.

At the moment, it works like this:
```cr
message.parse_mentions do |mention|
  case mention.type
  when .role?
    mention.id # 120571255635181568
  when .emoji?
    mention.name # "rubytaco"
  when .everyone?
    client.create_guild_ban( #...
  end
end
```
I can make more specific structs instead of the enum still if you like.
I'll look in to adding position and size; at the time I did this PR the use case (& crystal knowledge in general) wasn't completely clear in my head so I left it out to get more feedback first.

RX14:
Not every mention type has an name, and only some have an ID. It should really be more typesafe instead of a nillable munged thing idk.

z64:
I agree :+1:

RX14:
revert dis, we literally agreed that this needed changes

meew0:
whoops I thought the changes were already done, sorry
will revert

RX14:
This was, reverted, please PR again

PR #69, opened 2017-05-22
Remove REST#query_users, REST#accept_invite

z64:
- The query endpoint no longer exists.

- The scope of this library is not aimed at user bots, so I'm not sure why the accept invite endpoint was included.

### DIFF
--- a/src/discordcr/rest.cr
+++ b/src/discordcr/rest.cr
@@ -1091,22 +1091,6 @@ module Discord
       User.from_json(response.body)
     end
 
-    # Queries users by username.
-    #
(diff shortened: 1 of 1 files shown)


meew0:
Bots can use the accept invite endpoint during the authorisation process with a bearer token, and while it's true that we don't support user bots, we do support user operations using OAuth2.

z64:
So some usage may be like the following?
```ruby
client = Discord::Client.new token: "Bearer XYZ"
client.accept_invite("ABCdef")
```
You're right, that's acceptable then. I think I would at least like to expand the documentation a little to mention this.

PR #76, opened 2017-08-13
Fix status updates being mostly broken

### LINKED ISSUE #71: client.status_update does not seem to work
```  
  game = Discord::GamePlaying.new(name: "Test game")
  client.status_update(status: "idle", game: game, afk: false, since: 0_i64)
````

This code is not working at all. My bot is not "idle" and the game is not set. What am I doing wrong?

### PR DESCRIPTION
**Pull request by meew0**
Apparently Discord is very specific about the `since` field: if it is 0 it is treated differently from being null, causing unpredictable behaviour. Now it is null by default. Fixes #71.

### DIFF
--- a/src/discordcr/client.cr
+++ b/src/discordcr/client.cr
@@ -242,7 +242,7 @@ module Discord
     # *afk* can be used in conjunction to signify to Discord that the status
     # change is due to inactivity on the bot's part – this fulfills no cosmetic
     # purpose.
-    def status_update(status : String? = nil, game : GamePlaying? = nil, afk : Bool = false, since : Int64 = 0_i64)
+    def status_update(status : String? = nil, game : GamePlaying? = nil, afk : Bool = false, since : Int64? = nil)
       packet = Gateway::StatusUpdatePacket.new(status, game, afk, since)
       @websocket.send(packet.to_json)
     end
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/discordcr/client.cr [hunk 1]
```diff
@@ -242,8 +242,9 @@ module Discord
     # *afk* can be used in conjunction to signify to Discord that the status
     # change is due to inactivity on the bot's part – this fulfills no cosmetic
     # purpose.
-    def status_update(status : String? = nil, game : GamePlaying? = nil, afk : Bool = false, since : Int64 = 0_i64)
+    def status_update(status : String? = nil, game : GamePlaying? = nil, afk : Bool = false, since : Int64? = nil)
       packet = Gateway::StatusUpdatePacket.new(status, game, afk, since)
+      puts packet.to_json
```
RX14: deletos delet this

src/discordcr/mappings/gateway.cr
```diff
@@ -104,7 +104,7 @@ module Discord
         status: {type: String, nilable: true, emit_null: true},
         game: {type: GamePlaying, nilable: true, emit_null: true},
         afk: Bool,
-        since: Int64
+        since: {type: Int64, nilable: true, emit_null: true}
```
RX14: We should really replace all the `Int64, nilable: true` with `Int64?` but that should be a whole new PR.

PR #77, opened 2017-08-14
Update docs badge to use master instead of 0.1.0

g33kidd:


RX14:
We should track the latest release, not master.

RX14:
Should be `v0.3.0` and depends on #79

RX14:
@meew0 73 commits to master since the last release is it not time?

meew0:
I'll update the docs badge manually when the next release causes a doc update, see #139

Issue #60, opened 2017-05-08
Guild Members and Users not being cached


z64:
In events such as `GUILD_CREATE` (and others), discordcr will not populate `@users` and `@members` as expected.

---
discordrb illustrates the missing behavior on `GUILD_CREATE`:

1. `GUILD_CREATE` in the gateway `case` is called, which follows up with `Cache#ensure_server`.

https://github.com/meew0/discordrb/blob/master/lib/discordrb/cache.rb#L163

2. `#ensure_server` instances a new server, whose `#initialize` calls `Server#process_members`

https://github.com/meew0/discordrb/blob/master/lib/discordrb/data.rb#L2795-L2801

3. This calls `Member#initialize`, which caches the `User`

https://github.com/meew0/discordrb/blob/master/lib/discordrb/data.rb#L541

---

discordcr's cache behavior processes the guild, the roles, and the channels, but not members:

https://github.com/meew0/discordcr/blob/master/src/discordcr/client.cr#L387-L404

It would appear that it is very likely that `resolve_member` is almost always making a REST call for members we've could have already known about.

The `GuildMember` (or similar `User`-bearing class) `#cache` overload also doesn't cache the `User` object that `GuildMember` carries. Whether or not this is actually desired, I'm not sure.

https://github.com/meew0/discordcr/blob/master/src/discordcr/cache.cr#L152-L156

Minor note, but it would be nice to have getters for the `Cache`'s attributes for doing work across all of the cached data at once.

PR #78, opened 2017-08-14
Member cache fixes, and getters for cache hashes

meew0:
Fixes the two things mentioned in #60 — members and users from `GUILD_CREATE` not being cached, and missing getters for cache hashes. Also fixes a related bug uncovered during testing where nicks of cached members weren't updated properly.

### DIFF
--- a/src/discordcr/mappings/guild.cr
+++ b/src/discordcr/mappings/guild.cr
@@ -68,10 +68,22 @@ module Discord
 
   struct GuildMember
     # :nodoc:
-    def initialize(payload : Gateway::GuildMemberAddPayload | GuildMember, roles : Array(UInt64)? = nil)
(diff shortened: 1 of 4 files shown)


src/discordcr/mappings/guild.cr
```diff
@@ -59,10 +59,21 @@ module Discord
 
   struct GuildMember
     # :nodoc:
-    def initialize(payload : Gateway::GuildMemberAddPayload | GuildMember, roles : Array(UInt64)? = nil)
+    def initialize(payload : Gateway::GuildMemberAddPayload | GuildMember, roles : Array(UInt64), nick : String?)
+      # This redundant assignment is necessary because Crystal would otherwise
+      # erroneously assume @roles can be nil.
```
RX14: Please make an issue with this and add a TODO linking to the crystal issue before comitting this.

RX14: IIRC this is actually fixed in crystal now.

PR #79, opened 2017-08-14
Fix Travis tag builds being ignored for deployment

meew0:
Tag builds should of course be tested and deployed rather than just tested. Previously this case was missed during checking whether the deployment should be skipped.


RX14:
Have you, or is there a way to, test this?

RX14:
You'll likely have to deploy v0.3.0 tag to gh-pages manually.

PR #80, opened 2017-08-25
Fixed modify_* REST calls

zatherz:
Named tuples emit nulls for nil fields. The modify/PATCH endpoints expect the field to not exist at all to be omitted. This patch changes these calls to use a JSON.mapping struct instead, which by default does not emit nulls.

Additionally, missing `nil` defaults have been added too, to make calling these methods more convenient.

### DIFF
--- a/src/discordcr/mappings/rest.cr
+++ b/src/discordcr/mappings/rest.cr
@@ -15,5 +15,87 @@ module Discord
         pruned: UInt32
       )
     end
+
+    # A payload for the Modify Channel REST API call.
+    #
+    # Used instead of a tuple to control null emitting.
+    struct ModifyChannelPayload
+      JSON.mapping(
+        name: String?,
+        position: UInt32?,
+        topic: String?,
+        bitrate: UInt32?,
+        user_limit: UInt32?,
+        nsfw: Bool?
+      )
+
+      def initialize(@name, @position, @topic, @bitrate, @user_limit, @nsfw)
+      end
+    end
+
+    # A payload for the Modify Guild REST API call.
+    #
+    # Used instead of a tuple to control null emitting.
+    struct ModifyGuildPayload
+      JSON.mapping(
+        name: String?,
+        region: String?,
+        verification_level: UInt8?,
+        afk_channel_id: UInt64?,
+        afk_timeout: Int32?,
+        icon: String?,
+        owner_id: UInt64?,
+        splash: String?
+      )
+
+      def initialize(@name, @region, @verification_level, @afk_channel_id,
+                     @afk_timeout, @icon, @owner_id, @splash)
+      end
+    end
+
+    # A payload for the Modify Guild Member REST API call.
+    #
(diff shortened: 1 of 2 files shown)


examples/ping_with_response_time.cr
```diff
@@ -7,6 +7,7 @@ require "../src/discordcr"
 client = Discord::Client.new(token: "Bot MjI5NDU5NjgxOTU1NjUyMzM3.Cpnz31.GQ7K9xwZtvC40y8MPY3eTqjEIXm", client_id: 229459681955652337_u64)
 
 client.on_message_create do |payload|
+  client.modify_guild 0u64, name: "Hi"
```
greenbigfrog: Why is the guild name being modified here? (If this was added on purpose it should be moved in a extra example.)

zatherz: Great catch. Was a test to see if it'd compile. Going to fix & squash.

src/discordcr/mappings/rest.cr
```diff
@@ -15,5 +15,86 @@ module Discord
...
+    struct ModifyChannelPayload
+      JSON.mapping(
+        name: String?,
+        position: UInt32?,
+        topic: String?,
+        bitrate: UInt32?,
+        user_limit: UInt32?
+      )
+
+      def initialize(@name, @position, @topic, @bitrate, @user_limit)
```
z64: Looks like you added `nsfw` to the REST call but this `ModifyChannelPayload` object doesn't actually support it.

src/discordcr/rest.cr
```diff
@@ -577,20 +579,20 @@ module Discord
...
-    def modify_guild(guild_id : UInt64, name : String?, region : String?,
-                     verification_level : UInt8?, afk_channel_id : UInt64?,
-                     afk_timeout : Int32?, icon : String?, owner_id : UInt64?,
-                     splash : String?)
-      json = {
-        name:               name,
-        region:             region,
+    def modify_guild(guild_id : UInt64, name : String? = nil, region : String? = nil,
+                     verification_level : UInt8? = nil, afk_channel_id : UInt64? = nil,
+                     afk_timeout : Int32? = nil, icon : String? = nil, owner_id : UInt64? = nil,
```
z64: So, something I realized, is that if you actually want to *remove* a guilds icon (reset it to the default), in API terms you send:
```json
{
  "icon": null
}
```
which would mean passing `icon: nil` - but this would not get emitted by the JSON builder.

To actually remove the icon, this is also valid:
```json
{
  "icon": ""
}
```
Which we can do, and won't get wiped out by the builder via `icon: ""` in the method call.

While this is still fine (discarding nulls > strict arguments) it might be worth documenting this in case someone actually wants to do this, as it's not obvious IMO. Maybe with a `NOTE:` docstring.

This also applies to a user's `avatar`.

z64:
This is superseded by #94

PR #83, opened 2017-08-25
Add missing User fields

zatherz:
Some fields which are only available for `/users/@me`. [Mostly taken from documentation](https://discordapp.com/developers/docs/resources/user#user-object-user-structure), the `phone` field found through messing around with Postman
image

### DIFF
--- a/src/discordcr/mappings/user.cr
+++ b/src/discordcr/mappings/user.cr
@@ -20,8 +20,7 @@ module Discord
       email: {type: String, nilable: true},
       bot: {type: Bool, nilable: true},
       mfa_enabled: {type: Bool, nilable: true},
-      verified: {type: Bool, nilable: true},
-      phone: {type: String, nilable: true}
+      verified: {type: Bool, nilable: true}
     )
   end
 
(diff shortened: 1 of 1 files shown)


z64:
Why add `phone` if it will always be `nil`? I don't think Discord plans on exposing `phone` via OAuth2 any time soon, if at all.

zatherz:
The screenshot demonstrates `phone` being present in the result of `/users/@me`

src/discordcr/mappings/user.cr
```diff
@@ -18,7 +18,10 @@ module Discord
       discriminator: String,
       avatar: {type: String, nilable: true},
       email: {type: String, nilable: true},
-      bot: {type: Bool, nilable: true}
+      bot: {type: Bool, nilable: true},
+      mfa_enabled: {type: Bool, nilable: true},
+      verified: {type: Bool, nilable: true},
+      phone: {type: String, nilable: true}
```
meew0: As @z64 said, please remove the phone field

PR #84, opened 2017-08-28
Adds REST methods for guild emojis

zatherz:
* `REST#get_guild_emojis`
* `REST#modify_guild_emoji`
* `REST#create_guild_emoji`

### DIFF
--- a/src/discordcr/rest.cr
+++ b/src/discordcr/rest.cr
@@ -620,6 +620,59 @@ module Discord
       Guild.from_json(response.body)
     end
 
+    # Gets a list of emojis. Will not work for bots (it'll return an empty array).
+    #
+    # There are no available API docs for this method.
+    def get_guild_emojis(guild_id : UInt64)
+      response = request(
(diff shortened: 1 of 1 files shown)


src/discordcr/rest.cr
```diff
@@ -620,6 +620,59 @@ module Discord
       Guild.from_json(response.body)
     end
 
+    # Gets a list of emojis. Will not work for bots (it'll return an empty array).
```
z64: You should clarify these endpoints only work for whitelisted OAuth2 apps.

z64: Since this PR opened, managing emoji is now a public API :tada: 

You can adjust the docstrings to reflect this (it *will* work for bots). Now however, it is only the `roles` field of an emoji that requires whitelisting to create/modify.

zatherz: Don't have time to do this right now, if somebody does then go ahead and I'll close this PR

PR #85, opened 2017-08-28
Adds REST methods for vanity URL

zatherz:
* `REST#get_guild_vanity_url`
* `REST#modify_guild_vanity_url`


zatherz:
Tell me if `get_guild_vanity_url` should return the entire payload and not just the `code` which is the only field in that payload.

PR #87, opened 2017-08-29
Add missing GuildMemberUpdatePayload `nick` field

zatherz:


meew0:
Already implemented in #78

PR #91, opened 2017-09-01
Implement Webhook resource

azah:
Add REST routes for webhooks
Add ModifyWebhookPayload and ExecuteWebhookPayload to handle nil
Create webhook struct mapping in separate file
Check for response on execute_webhook if wait is true

### DIFF
--- a/src/discordcr/mappings/rest.cr
+++ b/src/discordcr/mappings/rest.cr
@@ -22,5 +22,54 @@ module Discord
         code: String
       )
     end
+
+    # A request to the Modify Webhook REST API call.
+    class ModifyWebhookPayload
+      JSON.mapping(
+        name: String?,
+        avatar: String?,
+        channel_id: UInt64?
+      )
+
+      def initialize(@name, @avatar, @channel_id)
+      end
+    end
+
+    # A request to the Execute Webhook REST API call.
+    class ExecuteWebhookPayload
+      JSON.mapping(
+        content: String?,
+        file: String?,
+        embeds: Array(Embed)?,
+        tts: Bool?,
+        avatar_url: String?,
+        username: String?
+      )
+
+      def initialize(@content, @file, @embeds, @tts, @avatar_url, @username)
+      end
+    end
+
+    # A response to the Modify Webhook REST API call.
+    class ModifyWebhookPayload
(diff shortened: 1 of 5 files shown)


z64:
Would you be interested in implementing the gateway event for webhooks too?

Docs: [`WEBHOOKS_UPDATE`](https://discordapp.com/developers/docs/topics/gateway#webhooks-update)

To be added to [`Client#handle_dispatch`](https://github.com/meew0/discordcr/blob/master/src/discordcr/client.cr#L297)

src/discordcr/mappings/rest.cr
```diff
@@ -15,5 +15,27 @@ module Discord
...
+
+    # A response to the Modify Webhook REST API call.
+    class ModifyWebhookPayload
+      JSON.mapping(name: String?, avatar: String?, channel_id: UInt64?)
+
+      def initialize(@name, @avatar, @channel_id)
+      end
+    end
+
+    class ExecuteWebhookPayload
```
meew0: Is there any point to having a particular mapping class? It only ever seems to be used once and is never exposed to the user.

src/discordcr/mappings/rest.cr
```diff
@@ -15,5 +15,27 @@ module Discord
         pruned: UInt32
       )
     end
+
+    # A response to the Modify Webhook REST API call.
+    class ModifyWebhookPayload
+      JSON.mapping(name: String?, avatar: String?, channel_id: UInt64?)
```
meew0: Please make the formatting here consistent with all the other `JSON.mapping` calls (newlines)

src/discordcr/mappings/rest.cr
```diff
@@ -15,5 +15,27 @@ module Discord
         pruned: UInt32
       )
     end
+
+    # A response to the Modify Webhook REST API call.
```
meew0: Should be "request", not "response".

src/discordcr/mappings/rest.cr
azah: How should it be changed?

azah:
travis: `./deploy.sh: line 15: [: missing `]'`

PR #92, opened 2017-09-02
[WIP] Refactor Cache

### LINKED ISSUE #31: Remove Cache from Client

### LINKED ISSUE #32: Make Cache abstract

z64:
# :warning: **WIP** :warning: 

Alright, I think this is at a point where I've kind of implemented all the concepts I wanted to capture with this cache to open a PR for any other early feedback.

## Current `Cache` Observations, Motivation

- Stores everything it can, cannot be customized without re-implementing or monkey patching
- Is always internal and in-memory
- Caching behavior is intertwined with gateway code
- Closes #31 
- Closes #32 

## Redesign

- Describes a common caching interface for a Discord type, `TypeCache`
- Users can write their own implementation of `TypeCache` and describe how they want the client to operate on them by defining `init_handlers`. For common Discord types, template collections of event handlers are provided for convenience which users can `include` on their own `TypeCache`, and extend by overriding `init_handlers` and calling `super`
- Introduces `CachedClient` which any `TypeCache` you like can be attached to. Inherits & overrides `Client`'s `REST` methods to cache `REST` results. Caches are managed by `CacheSet` which maps Discord types to their expected `TypeCache` generic.
- Provides a stock set of caches that implement `Hash` storage

The objective is to give users the tools necessary to implement whatever caching backend they like, with a minimal but extendable interface and behavior, chuck it inside `CachedClient`, and it should Just Work:tm:

### Usage

This currently compiles:
```cr
client = Discord::CachedClient.new(token: "token")

# Attach some Memory caches to `Guild` and `GuildMember` related events and REST calls
client.cache_set.configure(
  guild: Discord::Cache::Memory::Guild,
  guild_member: Discord::Cache::Memory::GuildMember
)

# Make cached calls
client.get_guild(225375815087554563_u64)

client.run
```

## Todo

- [ ] Implement the remaining `Handlers` and `Memory` caches
- [ ] Implement the remaining cache-able resources (`REST` and gateway delegations) into `CachedClient`
- [x] Implement some sort of relationship caching (ex: Guild ID -> Channel ID)? Redis is a personal implementation target (as a seperate shard) and sets would be a good use for this, but I'm not entirely sure about other implementations. I could use `Set`, but this would not work with `TypeCache` as is.
- [ ] Decide if iterators like `#each` should be expected from `TypeCache` or users should just implement that on their own. Perhaps only on the `Memory` caches
- [ ] A shortcut for spinning up `CachedClient` with all `Memory` caches (constant tuple or something)
- [ ] Separate logger for caching? Remove or lower debugging logging level either way. Probably log more things too (handler calls?)
- [ ] More docs in places could be used, I'm sure
- [ ] Format
- [ ] Testing
- [ ] Tidy commit history

### Other thoughts

- The current macros work (perhaps they won't as I add more cache types) but I'm sure they could be tidied in some way that I couldn't figure out yet. Feedback especially welcome here.

### DIFF
--- a/src/discordcr/cache.cr
+++ b/src/discordcr/cache.cr
@@ -1,216 +1,241 @@
 require "./mappings/*"
 
-module Discord
-  # A cache is a utility class that stores various kinds of Discord objects,
-  # like `User`s, `Role`s etc. Its purpose is to reduce both the load on
-  # Discord's servers and reduce the latency caused by having to do an API call.
-  # It is recommended to use caching for bots that interact heavily with
-  # Discord-provided data, like for example administration bots, as opposed to
-  # bots that only interact by sending and receiving messages. For that latter
-  # kind, caching is usually even counter-productive as it only unnecessarily
-  # increases memory usage.
+# Caches are utility classes that stores various kinds of Discord objects, like `User`s, `Roles`, etc.
+# Its purpose is to reduce both the load on Discord's servers and reduce the latency caused by having
+# to do an API call. It is recommended to use caching for bots that interact heavily with Discord-provided
+# data, like for example administration bots, as opposed to bots that only interact by sending and receiving
+# messages. For that latter kind, caching is usually even counter-productive as it only unnecessarily
+# increases memory usage.
+#
+# This modules contains a basic in-memory cache, but also provides interfaces for you to build your own
+# custom cache class (see `TypeCache`) that may implement any other kind of store such as Redis or SQL ORDBMS.
+module Discord::Cache
+  # `TypeCache` is a generic interface that describes behavior for the storage of Discord-related objects.
+  # `TypeCache` behavior stores an object `V` by some unique identifier `K` (typically a UInt64 snowflake). 
+  # It is up to the user to implement the desired storage behavior.
+  # In the implementation of `#init_handlers`, the `TypeCache` is expected to register event handlers on the `Client`
+  # reference that will trigger cache-related actions to keep its resources up to date.
+  # Alternatively, the `TypeCache` can be used manually without or in addition to handling `Client` events.
   #
-  # Caching can either be used standalone, in a purely REST-based way:
+  # A simple implementation of an in-memory `Guild` cache may look like:
   # ```
-  # client = Discord::Client.new(token: "Bot token", client_id: 123_u64)
-  # cache = Discord::Cache.new(client)
+  # class GuildMemory
+  #   # Store guilds by their snowflake ID
+  #   include Discord::TypeCache(UInt64, Discord::Guild)
   #
-  # puts cache.resolve_user(66237334693085184) # will perform API call
-  # puts cache.resolve_user(66237334693085184) # will not perform an API call, as the data is now cached
-  # ```
+  #   # Include the template set of event handlers for guild-related events
+  #   include Discord::Handlers::Guild
   #
-  # It can also be integrated more deeply into a `Client` (specifically one that
-  # uses a gateway connection) to reduce cache misses even more by automatically
-  # caching data received over the gateway:
-  # ```
-  # client = Discord::Client.new(token: "Bot token", client_id: 123_u64)
-  # cache = Discord::Cache.new(client)
-  # client.cache = cache # Integrate the cache into the client
-  # ```
+  #   @guilds = {} of UInt64 => Discord::Guild
   #
-  # Note that if a cache is *not* used this way, its data will slowly go out of
-  # sync with Discord, and unless it is used in an environment with few changes
-  # likely to occur, a client without a gateway connection should probably
-  # refrain from caching at all.
-  class Cache
-    # Creates a new cache with a *client* that requests (in case of cache
-    # misses) should be done on.
-    def initialize(@client : Client)
-      @users = Hash(UInt64, User).new
-      @channels = Hash(UInt64, Channel).new
-      @guilds = Hash(UInt64, Guild).new
-      @members = Hash(UInt64, Hash(UInt64, GuildMember)).new
-      @roles = Hash(UInt64, Role).new
-
-      @dm_channels = Hash(UInt64, UInt64).new
-
-      @guild_roles = Hash(UInt64, Array(UInt64)).new
-      @guild_channels = Hash(UInt64, Array(UInt64)).new
+  #   def cache(id : UInt64, guild : Discord::Guild)
+  #     @guilds[id] = guild
+  #   end
+  #
+  #   def resolve?(id : UInt64)
+  #     @guilds[id]?
+  #   end
+  #
+  #   def remove(id : UInt64)
+  #     @guilds.delete(id)
+  #   end
+  # end
+  #
+  # cache = GuildMemory.new(client)
+  # ```
+  module TypeCache(K, V)
+    # Exception to raise when a member of a cache isn't found
+    class MissingMember < Exception
     end
 
-    # Resolves a user by its *ID*. If the requested object is not cached, it
-    # will do an API call.
-    def resolve_user(id : UInt64) : User
-      @users.fetch(id) { @users[id] = @client.get_user(id) }
+    def initialize(@client : Discord::Client)
+      init_handlers
     end
 
-    # Resolves a channel by its *ID*. If the requested object is not cached, it
-    # will do an API call.
-    def resolve_channel(id : UInt64) : Channel
-      @channels.fetch(id) { @channels[id] = @client.get_channel(id) }
+    # Stores the given `object` at `key`.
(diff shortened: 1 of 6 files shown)


unleashy:
they grow so fast

src/discordcr/cache.cr
```diff
@@ -1,216 +1,241 @@
...
+
+    def initialize(@client)
+    end
+
+    # The different possible cache types
+    TYPES = {
+      guild:        {key: UInt64, type: Guild},
+      guild_member: {key: {UInt64, UInt64}, type: GuildMember},
+      channel:      {key: UInt64, type: Channel},
+      role:         {key: {UInt64, UInt64}, type: Role},
```
meew0: Roles are uniquely identified by their ID and don't need the guild ID like members do. Instead, there should be a separate cache type that stores guild-role associations (and others for guild-channel associations, etc.)

src/discordcr/cache/guild.cr
```diff
@@ -0,0 +1,93 @@
...
+      end
+
+      def remove(id : UInt64)
+        Discord::LOGGER.info("Removing guild: #{id} (#{@guilds.size})")
+        @guilds.delete(id)
+      end
+    end
+
+    class GuildMember
+      # Store guilds by their snowflake ID
```
meew0: guild members

src/discordcr/cache/guild.cr
```diff
@@ -0,0 +1,93 @@
...
+      include TypeCache({UInt64, UInt64}, Discord::GuildMember)
+
+      # Include the template set of event handlers for guild-related events
```
meew0: guild member-related

src/discordcr/cache/guild.cr
```diff
@@ -0,0 +1,93 @@
...
+
+      def remove(id : {UInt64, UInt64})
+        Discord::LOGGER.info("Removing guild_member: #{id}")
+        @guild_members_map[id[0]].delete(id[1])
+        @guild_members.delete(id)
+      end
+    end
+
+    class Role
+      # Store guilds by their snowflake ID
```
meew0: roles

src/discordcr/cache/guild.cr
```diff
@@ -0,0 +1,93 @@
...
+      include TypeCache({UInt64, UInt64}, Discord::Role)
+
+      # Include the template set of event handlers for guild-related events
```
meew0: role-related

src/discordcr/cache.cr
```diff
@@ -1,216 +1,241 @@
...
+      users:        {key: UInt64, type: User},
+    }
+
+    {% begin %}
```
meew0: What is the point of this begin/end block? It seems redundant to me.

src/discordcr/cache.cr
```diff
@@ -1,216 +1,241 @@
...
+      {% for key, properties in TYPES %}
+        # Cache for `{{properties[:type]}}` objects
+        property {{key}} : TypeCache({{properties[:key]}}, {{properties[:type]}})?
+
+        def [](type : {{properties[:type]}}.class)
```
meew0: This method is ambiguous if there are multiple cache types with the same `type` property (e.g. guild-role ID mapping and guild-channel ID mapping)

meew0:
Some general comments:

- I think `CachedClient` is a bad idea. Hiding the caching logic in methods that usually do REST requests makes the whole thing very opaque. If a user wants to just call a REST method without potentially getting a cached result, they have no option to do so. If a user gets a wrong result they now have to consider library bugs in addition to Discord ones. It also means all the REST methods have to be duplicated, so if someone changes the REST methods for whatever reason but forgets to update the `CachedClient` ones, there will be a discrepancy. Furthermore, this duplication has to occur on a client class, which defeats the whole point of having the REST methods in a separate module. Above all though, it means clients and caches will now be as intertwined as they were before—separating this logic was the entire purpose of #31 and #32.
- Without `CachedClient`, is there any point to `CacheSet`?
- As for your question on iterators: I think iterators are an essential part of caching, because they allow for more detailed queries—getting a user by their name, for example, would impossible currently unless you specifically define such a method on the relevant `TypeCache`.

src/discordcr/cache.cr
z64: This is intentional ; I wasn't planning on "allowing" multiple stores for a single type. This method is really here to simplify the `cached_route` macro in `CachedClient`, where I can pass the route's argument type instead of transforming the AST to call the `property {{key}}`.

Anyways; are multiple stores of a given type really desired? I can't see why anyone would store things, for example, both in-memory *and* in a database.

src/discordcr/cache.cr
z64: The purpose of the "compound" tuple keys here is that they reflect the REST route argument types required to resolve them from the API.

Having a cache for `role: {key: UInt64, type: Role}` and something like `guild_roles: {key: UInt64, type: UInt64}` is also a valid design, but it also complicates setting up the cache / designing your own. Some may not even need this, as its only really needed if you plan on operating across the whole set of cached objects.

At the moment I've opted to leave this up to the user, and provide a second `resolve` method in the stock `Memory` caches that takes only the first ID (the guild ID) and access a second in-memory store that manages that relationship.

https://github.com/meew0/discordcr/pull/92/files/3a7578251e06aa6f28dc35d00b61e0747ea3d095#diff-a5ef42fdb346fff1caf1cb1e65c2cc16R49

Does that make sense? We can still go that way (separate relationship store).

z64:
Thanks for the feedback so far.

>  If a user wants to just call a REST method without potentially getting a cached result, they have no option to do so.

I think @RX14 mentioned at one point additionally providing `uncached_{{rest_method}}` to have this behavior. This would be easy to generate alongside the cached route ones, which should be the default.

> Without CachedClient, is there any point to CacheSet?

I may get rid of this still and move it into `CachedClient`. However, if someone wanted to operate the cache manually (without a `Client`, for whatever reason), or any other kind of exotic client, it provides a common cache container for them to use. But that's probably not a strong enough argument for keeping it around.

> As for your question on iterators:

Agreed.

src/discordcr/cache.cr
z64: I don't fully understand the issue, but without wrapping this in a `begin..end` macro, it doesn't compile. It's been a week or so since I looked at this, I'll double check.

RX14:
I think that actually an uncached optional parameter would be best for the methods. This should be implemented and ignored on the base client so they have the same interface.

z64:
I'm revisiting this now, and after my experiences with discordcr-middleware, I think we could take a much simpler approach to this:

```cr
record User, id : UInt64, name : String

# The Discord API :TM:
DISCORD = {
  1_u64 => User.new(1_u64, "z64"),
  2_u64 => User.new(2_u64, "PixeL"),
}

# REST mixin
module REST
  def get_user(id : UInt64)
    puts "Making REST request!"
    DISCORD[id]
  end
end

# Client that implements T as a cache
class CachedClient(T)
  include REST

  getter cache : T

  def initialize(@cache : T = T.new)
  end

  def get_user(id : UInt64, cached : Bool = true)
    if cached
      @cache.resolve_user(id) { super(id) }
    else
      @cache.cache super(id)
    end
  end
end

# MemoryCache is now expected to implement whatever Client asks it to
class MemoryCache
  getter users = {} of UInt64 => User

  def cache(user : User)
    @users[user.id] = user
  end

  # Passes back up to the client on a miss
  def resolve_user(id)
    @users.fetch(id) { @users[id] = yield }
  end
end

client = CachedClient(MemoryCache).new
p client.get_user(1_u64) # Makes a REST request
p client.get_user(1_u64) # Pull from cache

# Cache something ahead of time, i.e from WS
client.cache.cache DISCORD[2_u64]
p client.get_user(2_u64)                # Pull from cache
p client.get_user(2_u64, cached: false) # Makes a REST request
```
By doing this:

1. You give `CachedClient` a `T`, and instead of creating some abstract contract, the compiler will ask you to implement whatever caching methods `CachedClient` wants to use.
2. Controlling what types get cached and what doesn't is as simple as adding a method that explicitly does nothing.
3. It's fully customizable via code, or you can create you own means of customization in `T#initialize`

However, in addition, I think this change could be simplified further: (which i'm 50/50 on, but I think it makes sense)
```cr
class CachedClient(T)
  # The full client, as it exists now, but with the generic T
end

# Recreate Client as an alias with a cache that does nothing
alias Client = CachedClient(NullCache)
```
All of the above combined would be a much smaller diff that enables users to use custom stores in a much less hacky/convoluted way that I've started here.

Does this seem like a better way forward? Other things like removing caching logic from the dispatch loop (if we still want to do that, I'm not sure there's a point now) can be split into later PRs this way too.

PR #94, opened 2017-10-17
Add helper method for encoding tuples to JSON, omitting pairs with null values

z64:
This accomplishes the same (supersedes) as #80 , by making a single helper method that encodes tuples using `JSON.build` that skips null-valued keys for the same performance, as far as I can see.

While there's nothing wrong with additional structs in #80, if the *only* reason there are there is to control null-emission when building JSON, it is much less work to maintain the REST binding with this as you no longer have to maintain *both* the methods and the structs as API methods are added or fixed - you just call `encode_tuple` and don't have to think about anything else.

I originally wanted to make this a private method, as there's no real reason to expose it, but I'm not sure how I would spec the private method, as that doesn't compile.

This also does some small housekeeping of removing setting of temporary `response =` variables when they weren't being used at all in the rest of the method.

I originally wasn't going to open this, but I had it laying around, #80 still wasn't merged, and #HACKTOBER Y'ALL

ahem

### DIFF
--- a/src/discordcr/rest.cr
+++ b/src/discordcr/rest.cr
@@ -96,6 +96,18 @@ module Discord
       response
     end
 
+    # :nodoc:
+    def encode_tuple(**tuple)
+      JSON.build do |builder|
+        builder.object do
+          tuple.each do |key, value|
+            next if value.nil?
+            builder.field(key) { value.to_json(builder) }
+          end
+        end
+      end
+    end
+
     # Gets the gateway URL to connect to.
     #
     # [API docs for this method](https://discordapp.com/developers/docs/topics/gateway#get-gateway)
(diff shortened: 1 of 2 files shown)


src/discordcr/rest.cr
```diff
@@ -132,15 +144,16 @@ module Discord
...
-        name:       name,
-        position:   position,
-        topic:      topic,
-        bitrate:    bitrate,
-        user_limit: user_limit,
-      }.to_json
+    def modify_channel(channel_id : UInt64, name : String? = nil, position : UInt32? = nil,
+                       topic : String? = nil, bitrate : UInt32? = nil, user_limit : UInt32? = nil,
+                       nsfw : Bool? = nil)
+      json = encode_tuple(
```
RX14: This can't compile, because you're not passing a namedtuple into `encode_tuple`.

z64: I had fixed but hadn't pushed it ;(

Is the double-splat acceptable or would you actually prefer the restrictions & pass named tuple literals? I'm not fully aware of how the compiler types `**args` / what difference it would make.

meew0:
Can you rebase this?

z64:
Yep 👍

z64:
Done; though CI is failing because master is broken.

RX14:
needs a rebase

z64:
Done

PR #95, opened 2017-10-22
Fixed exception in get_current_user_guilds

hedgehog1029:
Icon is a nilable property - if a server has no icon, the value will be null.
This patch fixes that by setting `icon` as a nilable JSON property.

### DIFF
--- a/src/discordcr/rest.cr
+++ b/src/discordcr/rest.cr
@@ -1166,11 +1166,11 @@ module Discord
       path = "/users/@me/guilds?limit=#{limit}"
 
       if before > 0
-          path += "&before=#{before}"
+        path += "&before=#{before}"
       end
 
       if after > 0
-          path += "&after=#{after}"
+        path += "&after=#{after}"
       end
 
       response = request(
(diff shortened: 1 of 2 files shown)


hedgehog1029:
I've just fixed another small issue in `Discord::REST#get_guild_channels` where the URL was trying to reference a non-existent `channel_id` variable, where instead it should be `guild_id`.

I've also added support for pagination on `get_current_user_guilds` - it follows the API's recommended defaults. If there are any required changes before this can be merged I'll happily implement them.

RX14:
Why is this 1 PR with unrelated changes? Each unrelated change should be in a seperate PR.

hedgehog1029:
The only unrelated part is the one-line change in c7ae47f - but I felt that it would be silly to create an entire PR for a one-line change.

src/discordcr/rest.cr
```diff
@@ -1162,12 +1162,22 @@ module Discord
...
-    def get_current_user_guilds
+    def get_current_user_guilds(limit : UInt8 = 100, before : UInt64 = 0, after : UInt64 = 0)
+      path = "/users/@me/guilds?limit=#{limit}"
+
+      if before > 0
+          path += "&before=#{before}"
+      end
+
+      if after > 0
+          path += "&after=#{after}"
```
RX14: Please use `HTTP::Params` to build this.

RX14:
@meew0 fix CI pls tho

PR #96, opened 2017-11-05
Fix mismatches of default arguments type with their restrictions

z64:
These methods would fail to compile with: "can't restrict Int32 to UInt32"


RX14:
💢 who PRed this without testing 💢

PR #99, opened 2017-11-16
Use TimestampConverter instead of DATE_FORMAT

z64:
Discord sometimes sends timestamps without floating point accuracy, which is still a valid ISO 8601 timestamp. This would throw an exception and fail to parse the message.

### DIFF
--- a/spec/discordcr_spec.cr
+++ b/spec/discordcr_spec.cr
@@ -19,11 +19,33 @@ struct StructWithSnowflakeArray
   )
 end
 
+struct StructWithTime
+  JSON.mapping(
+    data: {type: Time, converter: Discord::TimestampConverter}
+  )
+end
+
 describe Discord do
   describe "VERSION" do
     it "matches shards.yml" do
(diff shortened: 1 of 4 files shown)


spec/discordcr_spec.cr
```diff
@@ -27,6 +27,16 @@ describe Discord do
     end
   end
 
+  describe Discord::DATE_FORMAT do
+    it "parses a time with floating point accuracy" do
+      Discord::DATE_FORMAT.parse "2017-11-16T13:09:18.291000+00:00"
```
RX14: This parses but only preserves per-second accuracy.

RX14:
The ISO8601 support in cr suucks

z64:
Haha. Well, do we write a more complicated converter then? The current `DATE_FORMAT` means you can't handle some messages at all because they fail to be instantiated.

RX14:
Literally trying to parse with fraction then revering to non-fractional would be the simplest idea.

RX14:
rebase this to fix CI pls

z64:
Done

PR #102, opened 2017-11-24
Fix REST#modify_guild_channel_positions

z64:
This endpoint doesn't work as written. I've renamed it to match the intended endpoint, and added a struct to pass as an argument.

It may be helpful to add an abstraction that lets you move channels around in the most basic manner, as [it's kind of a pain](https://github.com/z64/discordrb/blob/8972d84b222bcdcd027e4f81e8dd9e71a0e57aaa/lib/discordrb/data.rb#L1374-L1434). Would this be acceptable? If so I'll do this in another PR. This one at least makes the endpoint usable.

### DIFF
--- a/src/discordcr/mappings/rest.cr
+++ b/src/discordcr/mappings/rest.cr
@@ -2,6 +2,13 @@ require "./converters"
 
 module Discord
   module REST
+    # Enum for controlling emission of JSON key/value pairs where
+    # JSON `null` is a significant value.
+    enum MaybeField
+      None
+      Unchanged
+    end
+
     # A response to the Get Gateway REST API call.
     struct GatewayResponse
       JSON.mapping(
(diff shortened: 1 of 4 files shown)


z64:
However, it seems that this works except for one case - when wanting to remove a channel from a category.

With this patch, you cannot send `"parent_id": null` because null values are not emitted. We *also* cannot do `emit_null: true` for `parent_id` because the key must be absent if we are *not* changing the channels parent, as the `null` value means "remove the parent". Additionally, only one channels category can be changed at a time with this endpoint.

The only solution I can think of is to have `ModifyGuildChannelPositionPayload` that doesn't have `parent_id` and `ModifyGuildChannelPositionWithParentPayload` that does have it. Does that sound like the thing to do?

RX14:
So `null` is treated differently to absent. We should abstract it into an `Enum` to represent the absent and null behaviours instead of overriding `nil` and write a custom converter.

z64:
Alrighty, I implemented the enum `REST::MaybeField` (taking better name suggestions..)

As I mentioned in chat, I don't see this as being possible with a converter since the key is always written first, so I've removed the `JSON.mapping` and written `def to_json(builder : JSON::Builder)` which will be called when it is serialized from within a collection (`Array(ModifyGuildChannelPositionPayload)`) that will start the document.

src/discordcr/mappings/rest.cr
```diff
@@ -2,6 +2,13 @@ require "./converters"
 
 module Discord
   module REST
+    # Enum for controlling emission of JSON key/value pairs where
+    # JSON `null` is a significant value.
```
z64: I could probably be more explicit here about how the enum is intended to be used.

"`MaybeField::None` will emit a key and a JSON `null` value. `MaybeField::Unchanged` will not emit a key or a value."

src/discordcr/mappings/rest.cr
```diff
@@ -22,5 +29,37 @@ module Discord
...
+            SnowflakeConverter.to_json(@id, builder)
+          end
+
+          builder.field("position", @position)
+
+          parent = @parent_id
+
+          if parent.is_a?(MaybeField)
+            if parent.none?
+              builder.field("parent_id", nil)
```
z64: Should I prefer a guard style here? I think that would be more consistent with the `lock_permissions` field `unless` guard below.

z64:
Updated with `ChannelParent`, etc.

If this is going to be the "style" going forward (and we're going to have many of these enums) I figured it would be best to keep them in `enums.cr` instead of mixed in with the structs and REST call methods.

meew0:
github web merge conflict resolver: never again

PR #105, opened 2017-12-01
Test converters directly, test serialization

z64:
The existing specs for converters weren't testing the converters directly. Additionally, we weren't testing serialization.

This refactor addresses this with a DRY spec helper that tests both parsing and serialization.

### DIFF
--- a/spec/converters_spec.cr
+++ b/spec/converters_spec.cr
@@ -1,0 +1,29 @@
+require "./spec_helper"
+
+def it_converts(json, to value, with converter, file = __FILE__, line = __LINE__)
+  describe converter, file, line do
+    it "converts #{json} to #{value.class}" do
+      parser = JSON::PullParser.new(json)
+      converted = converter.from_json(parser)
+      converted.should eq value
+    end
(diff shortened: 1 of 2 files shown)


RX14:
i'm fine with the new specs but i'm 50/50 on removing the old ones

z64:
How do you mean? These are the same specs refactored to not use a struct wrapper that didn't have anything to do with what was being tested; we're testing the converters, not JSON.mapping 
(there's no "new" specs here)

z64:
I've rebased this, but it shouldn't be accepted yet pending #130 as this deletes all of the snowflake converters that this PR was primarily made to address. It could maybe be made to address the other remaining converters, but I'll wait until #130 is taken care of before I revisit this.

z64:
This PR is now basically replaced by `snowflake_spec`, as `Snowflake` handles serializing and parsing itself. I still think the other converter specs should be refactored in a similar fashion, but I'll revisit that some other time. The converters themselves need some changes too to make them more consistent, I think.

PR #108, opened 2017-12-23
fix type for list_guild_members

greenbigfrog:
The number expected by Discord is `0-1000`. Also the default should be 1000, since most of the time one will want to get all members at once.

### DIFF
--- a/src/discordcr/rest.cr
+++ b/src/discordcr/rest.cr
@@ -805,7 +805,7 @@ module Discord
     # to specify what ID to start at.
     #
     # [API docs for this method](https://discordapp.com/developers/docs/resources/guild#list-guild-members)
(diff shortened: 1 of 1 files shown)


src/discordcr/rest.cr
```diff
@@ -805,7 +805,7 @@ module Discord
     # to specify what ID to start at.
     #
     # [API docs for this method](https://discordapp.com/developers/docs/resources/guild#list-guild-members)
-    def list_guild_members(guild_id : UInt64, limit : UInt8 = 1, after : UInt64 = 0)
+    def list_guild_members(guild_id : UInt64, limit : UInt16 = 1_u16, after : UInt64 = 0_u64)
```
greenbigfrog: @meew0 is there a reason, why you had unsigned ints in the first place here? shouldn't it be normal `Int32`s?

PR #110, opened 2018-01-09
welcome.cr fix

r00ster91:
I changed that its using a channel_id instead of a guild_id at creating the welcome message. I also added the ability to set the channel where the welcome message should get sended to with !setwelcome (channel).

I know that this welcome-bot is still improvable (for example that he doesn't takes everything as an argument at !setwelcome) but I want it to be simple and not too complex.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .DS_Store
 /doc/
+/docs/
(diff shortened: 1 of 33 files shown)


unleashy:
this is good

PR #111, opened 2018-01-09
Improve welcome.cr

r00ster91:
I fixed it by using a channel_id instead of a guild_id at creating the welcome message. I also added the ability to set the channel where the welcome message should get sended to with !setwelcome (channel).

### DIFF
--- a/examples/welcome.cr
+++ b/examples/welcome.cr
@@ -1,4 +1,4 @@
-# This simple example bot creates a message whenever a new user joins the server
+# This simple example bot creates a message whenever a new user joins the server.
 
 require "../src/discordcr"
 
@@ -7,11 +7,25 @@ client = Discord::Client.new(token: "Bot MjI5NDU5NjgxOTU1NjUyMzM3.Cpnz31.GQ7K9xw
 cache = Discord::Cache.new(client)
 client.cache = cache
 
+welcome_channel = 0_u64
+
+client.on_message_create do |payload|
+  # Here we take for example "#welcome", "#general" or some other channel as the argument and then we put the ID of it into welcome_channel.
+  if payload.content.starts_with? "!setwelcome"
+    welcome_channel = payload.content.split(" ")[1][2..-2].to_u64
+  end
+end
+
 client.on_guild_member_add do |payload|
-  # get the guild/server information
+  # Break up if the welcome_channel isn't set yet.
+  if welcome_channel==0
+    next
+  end
+
+  # Get the guild/server information.
   guild = cache.resolve_guild(payload.guild_id)
 
-  client.create_message(guild.id, "Please welcome <@#{payload.user.id}> to #{guild.name}.")
+  client.create_message(welcome_channel, "Please welcome <@#{payload.user.id}> to #{guild.name}.")
 end
 
 client.run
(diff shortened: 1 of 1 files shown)


examples/welcome.cr
```diff
@@ -7,11 +7,25 @@ client = Discord::Client.new(token: "Bot MjI5NDU5NjgxOTU1NjUyMzM3.Cpnz31.GQ7K9xw
 cache = Discord::Cache.new(client)
 client.cache = cache
 
+welcome_channel = 0_u64
+
+client.on_message_create do |payload|
+  # Here we take for example "#welcome", "#general" or some other channel as the argument and then we put the ID of it into welcome_channel.
+  if payload.content.starts_with? "!setwelcome"
+    welcome_channel = payload.content.split(" ")[1][2..-2].to_u64
```
z64: This is pretty unsafe. This should probably done with regex and return an error to the user if the argument is invalid.

examples/welcome.cr
```diff
@@ -7,11 +7,25 @@ client = Discord::Client.new(token: "Bot MjI5NDU5NjgxOTU1NjUyMzM3.Cpnz31.GQ7K9xw
 cache = Discord::Cache.new(client)
 client.cache = cache
 
+welcome_channel = 0_u64
```
z64: Just set this to `nil`.

examples/welcome.cr
```diff
@@ -7,11 +7,25 @@ client = Discord::Client.new(token: "Bot MjI5NDU5NjgxOTU1NjUyMzM3.Cpnz31.GQ7K9xw
...
+  # Here we take for example "#welcome", "#general" or some other channel as the argument and then we put the ID of it into welcome_channel.
+  if payload.content.starts_with? "!setwelcome"
+    welcome_channel = payload.content.split(" ")[1][2..-2].to_u64
+  end
+end
+
 client.on_guild_member_add do |payload|
-  # get the guild/server information
+  # Break up if the welcome_channel isn't set yet.
+  if welcome_channel==0
```
z64: With this temporary variable set as `nil`, you can just do in a single line:
```cr
next unless welcome_channel
```

examples/welcome.cr
r00ster91: Yes I know thats what I meant with "I know that this bot is still improvable (for example that he doesn't takes everything as an argument at !setwelcome) but I want it to be simple and not too complex.". I thought first you could maybe leave that to others because so they learn more. Because im very sure the most people who look at these examples are beginners. But okay I think I will change that.

z64: Validating user input is a core concept to making chat bots. Doing something simple with regex should be something everyone should be familiar with, and still digestible for beginners, granted, this library intentionally is not aimed at beginners to Discord's API.

But yeah, please do this. This is way too error prone, and if users mess up the input it will only make them more confused trying to address exceptions from cases that aren't covered here.

examples/welcome.cr
r00ster91: I set this to 0u64 because when its nil I get:
Error in welcome.cr:21: no overload matches 'Discord::Client#create_message' with types (UInt64 | Nil), String
Overloads are:
 - Discord::REST#create_message(channel_id : UInt64, content : String, embed : Embed | ::Nil = nil)

    client.create_message(welcome_channel, "Please welcome <@#{payload.user.id}> to #{guild.name}.")

z64: Because you need to reduce the type [with something like an if branch](https://carc.in/#/r/3d0t) as I [explained in my other comment](https://github.com/meew0/discordcr/pull/111#discussion_r160522796) so that `welcome_message` is only `UInt64` when you go to send the message.

meew0:
I don't think the example needs to be this complex. There is an issue though because Discord no longer has the concept of a fixed general channel, but I don't think the way to resolve that would be to add the extra commands, as the example is really just meant to show off that you can use events other than `on_message_create`.

Issue #89, opened 2017-08-29
Do more logging


RX14:
Log more stuff so it's easier to debug

PR #116, opened 2018-02-02
Additional REST and WebSocket logging

z64:
Ref #89, continues #113 

Logs:
- REST in/out, and rate limiting. Prints in/out bodies on DEBUG severity
- WS in/out
- Client messages identified by shard key

Todo:
- [x] ~~Probably token redaction. Unsure of where to place this, though.. suggestions? I imagine we'll have to sub inside of `Logger` but that doesn't sound fun since we don't know what client we're coming from. Will probably have to use some kind of token regex I guess.~~ After talking, users of discordcr should be fully aware of what the token is to not accidentally leak it somewhere.


z64:
I'm thinking we should change `Client` to accept an `@logger`, with one that defaults to STDOUT instead of a single `LOGGER`. That would simplify things and I think ultimately be more flexible for the user.

src/discordcr/rest.cr
```diff
@@ -34,6 +34,12 @@ module Discord
         mutexes[rate_limit_key].synchronize { }
         global_mutex.synchronize { }
 
+        if LOGGER.debug?
```
RX14: bad. log the info in info AND the body in debug (if it's enabled).

```cr
LOGGER.info "[HTTP OUT] #{method} #{path} (#{body.try &.size || 0} bytes)"
LOGGER.debug "[HTTP OUT] BODY: #{body}" if LOGGER.debug?
```

src/discordcr/rest.cr
```diff
@@ -52,8 +58,10 @@ module Discord
           end
 
           if response.headers["X-RateLimit-Global"]?
+            LOGGER.warn "Locking global mutex for #{retry_after}"
```
RX14: How about just s/Locking global mutex/Pausing requests/

src/discordcr/rest.cr
```diff
@@ -52,8 +58,10 @@ module Discord
...
             global_mutex.synchronize { sleep retry_after }
           else
+            LOGGER.warn "Locking #{rate_limit_key} mutex for #{retry_after} (#{response.status_code == 429 ? "too many requests" : "preemptively"})"
```
RX14: ditto

src/discordcr/rest.cr
```diff
@@ -65,6 +73,10 @@ module Discord
         end
       end
 
+      if response
+        LOGGER.debug "[HTTP IN] #{response.status_code} #{response.status_message} (#{response.body.size} bytes): #{response.body}"
```
RX14: ditto on the `info` and `debug` split?

src/discordcr/websocket.cr
```diff
@@ -23,6 +35,7 @@ module Discord
 
     def on_message(&handler : Packet ->)
       @websocket.on_message do |message|
+        LOGGER.debug "[WS IN] #{message}"
```
RX14: avoid throwing these strings away with `if LOGGER.debug?`

src/discordcr/websocket.cr
```diff
@@ -32,7 +45,12 @@ module Discord
       @websocket.on_close(&handler)
     end
 
-    delegate run, close, send, to: @websocket
+    delegate run, close, to: @websocket
+
+    def send(message)
+      LOGGER.debug "[WS OUT] #{message}"
```
RX14: ditto

src/discordcr/rest.cr
z64: I think its important for the user to know they've actually hit the global rate limit and thus all requests will be affected, vs only a single route being locked.

RX14:
need rebase

z64:
@RX14 Rebased

src/discordcr/client.cr
```diff
@@ -72,7 +72,9 @@ module Discord
                    @shard : Gateway::ShardKey? = nil,
                    @large_threshold : Int32 = 100,
                    @compress : Bool = false,
-                   @properties : Gateway::IdentifyProperties = DEFAULT_PROPERTIES)
+                   @properties : Gateway::IdentifyProperties = DEFAULT_PROPERTIES,
+                   @logger = Logger.new(STDOUT))
+      @logger.progname = "discordcr"
```
z64: Should I build the below `@client_name` into this? I pass the `@logger` to our `WebSocket` instance which won't have `shard` information. On the other hand, `WebSocket` only has debug level logging at the moment, which you probably wouldn't use anyways for a production/sharded application.

PR #121, opened 2018-02-22
Add Message#activity

Daniel-Worrall:
Adding activity as https://discordapp.com/developers/docs/resources/channel#message-object-message-activity-types


src/discordcr/mappings/channel.cr
```diff
@@ -16,7 +16,22 @@ module Discord
...
+    Join        = 1
+    Spectate    = 2
+    Listen      = 3
+    JoinRequest = 5
+  end
+
+  struct Activity
+    JSON.mapping(
+      type: ActivityType,
+      party_id: String
```
RX14: This field should be nillable.

PR #123, opened 2018-03-09
Add Mention.parse utility for parsing mentions

### LINKED ISSUE #35: Stuff for parsing mentions

z64:
Refactoring of #64 

I ended up liking this a lot better than imposing an API over `Message`, as there's times when your content may not be coming from Discord (database, cache). If we still want that (`Message#parse_mentions`) that is trivial to add.

From the feedback in #64, this has:
- Stronger typing; structs for each mention type instead of an enum, methods specific to each mention type.
- Each mention type has `#position` and `#size`
- A few specs
- Support for the new animated flag in custom emoji mentions

Notes:
- ~~I was unsure about the `#not_nil!` assertion on `Regex::MatchData#begin`, as I couldn't think of when in this implementation it would ever be nil.~~ N/A in parser refactor
- @RX14 , you had wanted `#user_mentions`, etc. - I missed this when I refactored it, but I thought it might be a bit much to wrap a simple `#select`, or I couldn't think of a use case where the block-case variant would work just as well. Let me know what you think.
- ~~Should the block-accepting `Mention.parse` still return an `Array(MentionType)`?~~ N/A in parser refactor

Closes #35

### DIFF
--- a/spec/mention_spec.cr
+++ b/spec/mention_spec.cr
@@ -1,0 +1,50 @@
+require "./spec_helper"
+
+def it_parses_message(string, into expected)
+  it "parses #{string.inspect} into #{expected}" do
+    parsed = Discord::Mention.parse(string)
+    parsed.should eq expected
+  end
+end
+
+describe Discord::Mention do
+  describe ".parse" do
+    it_parses_message(
+      "<@123><@!456>",
+      into: [
+        Discord::Mention::User.new(123_u64, 0, 6),
+        Discord::Mention::User.new(456_u64, 6, 7),
+      ]
+    )
+
+    it_parses_message(
+      "<@&123>",
+      into: [Discord::Mention::Role.new(123_u64, 0, 6)])
+
+    it_parses_message(
+      "<#123>",
+      into: [Discord::Mention::Channel.new(123_u64, 0, 6)])
+
+    it_parses_message(
+      "<:foo:123><a:bar:456>",
+      into: [
+        Discord::Mention::Emoji.new(false, "foo", 123_u64, 0, 10),
+        Discord::Mention::Emoji.new(true, "bar", 456_u64, 10, 11),
+      ]
+    )
+
+    it_parses_message(
+      "@everyone@here",
+      into: [
+        Discord::Mention::Everyone.new(0),
+        Discord::Mention::Here.new(9),
+      ]
+    )
+
+    context "with invalid mentions" do
+      it_parses_message(
+        "<<@123<@?123><#123<:foo:123<b:foo:123><@abc><@!abc>",
+        into: [] of Discord::Mention)
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


spec/mention_spec.cr
```diff
@@ -0,0 +1,38 @@
+require "./spec_helper"
+
+describe Discord::Mention do
+  string = "foo bar <@123> <@!123> <@&123> <#123> <:foo:123> <a:bar:123> @everyone @here"
+
+  describe ".parse" do
+    it "parses mentions" do
+      parsed = Discord::Mention.parse(string)
+
+      parsed.size.should eq 8
```
RX14: I'd rather you just constructed the expected result array and use `parsed.should eq expected`

src/discordcr/mention.cr
```diff
@@ -0,0 +1,109 @@
...
+      5
+    end
+  end
+
+  alias MentionType = User | Role | Channel | Emoji | Everyone | Here
+
+  # Returns an array of mentions found in a string
+  def self.parse(string : String)
+    results = [] of MentionType
+    MENTION_REGEX.each do |type, regexp|
```
RX14: How TF does this work, surely you'll get the results in the wrong order? Also regexes suck please parse properly it's not actually that hard.

z64: It was in my last implementation and wasn't commented on, so I didn't change it. Didn't think it would matter that much for 2000 character strings either. I'll change this to a parser then.

z64:
Alrighty. I spent some time looking at the compilers lexer and tried to throw something similar together that fits our much simpler needs here. Also my first time writing something like this. I broke out a spec helper method so that test cases can easily be added later if new markdown is added to Discord.

z64:
I also wrote a basic usage example that prints all the mentions found in the message. I can add this to the PR as well if we think it would be helpful to have: https://gist.github.com/z64/cf09ff0f5e6c4c1f1f9840d3c3791265

z64:
@RX14 thanks!

Just a few things to follow up on before I'm cool with this being merged:
- Did you think the example in the gist I linked in my last comment is worth including in `examples/`?
- In the other thread you mentioned wanting something like `#user_mentions`, etc for each mention type. Personally I think its easy enough to `#select` from the array / use `case` with the type they want.
- Should we have `Message#parse_mentions` like my original PR had that wraps `Mention.parse`? I'm thinking no.

RX14:
Yeah that example is nice, and sure just select `is_a?` should work.

meew0:
I like this approach, thanks! Not a huge fan of the example as it's kinda pointless in its functionality, but that's an unimportant issue.

PR #124, opened 2018-03-23
Setup heartbeats in #run, not #initialize

azah:
GET /gateway/bot requires authorization, so
for sharding, one has to create a throwaway
client to authenticate and get the recommended
number of shards.

This commit prevents heartbeat setup from
actually running until needed.

### DIFF
--- a/src/discordcr/client.cr
+++ b/src/discordcr/client.cr
@@ -27,6 +27,9 @@ module Discord
     # a voice client, for example
     getter session : Gateway::Session?
 
+    # Used if operating in sharded mode.
+    property shard : Gateway::ShardKey?
+
     @websocket : Discord::WebSocket
     @backoff : Float64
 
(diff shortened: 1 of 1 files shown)


RX14:
How does this help you? There's still no way to set `@shards` after initialize so it amkes no difference.

azah:
Rip. Well, then perhaps a `shards=` setter should be added?

RX14:
Yeah, and the default shard tuple should be `{1, 1}`, to avoid the nilable.

If discord for whatever reason doesn't treat `{1, 1}` the same as not sharding then we can solve that at a lower layer.

z64:
It's `{index, total}`. `{1, 1}` isn't a valid shard key.

azah:
Changed the default val to 
```cr
@shard : Gateway::ShardKey? = {shard_id: 1, num_shards: 1},
```
Which works. However, changing the shards using `shard=` does not work, it gives invalid shard.

RX14:
whats the point of making it nilable then

RX14:
Thats actually a pretty good idea

PR #131, opened 2018-05-02
Fix REST#add_guild_member when user is already a member

z64:
The API returns `204 No Content` if the user is already a member of the guild. (This isn't documented in the API docs :disappointed: )

### DIFF
--- a/src/discordcr/rest.cr
+++ b/src/discordcr/rest.cr
@@ -871,7 +871,11 @@ module Discord
         json
       )
 
-      GuildMember.from_json(response.body)
+      if response.status_code == 201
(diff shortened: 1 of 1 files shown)


src/discordcr/rest.cr
```diff
@@ -871,7 +871,7 @@ module Discord
         json
       )
 
-      GuildMember.from_json(response.body)
+      GuildMember.from_json(response.body) if response.status_code == 201
```
RX14: please don't use implicit nil here; be more explicit:
```cr
if response.status_code == 201
  GuildMember.from_json(response.body)
else
  nil
end
```

PR #136, opened 2018-05-10
Reimplement Cache

### LINKED ISSUE #31: Remove Cache from Client

### LINKED ISSUE #32: Make Cache abstract

### LINKED ISSUE #33: Allow configuration of GenericCache
Allow changing whether the cache caches each type.

z64:
Supersedes #92 
Closes #31 
Closes #32 
Closes #33 

- Can cache configuration be improved?
- Should we provide an alias for each cache type? It might be easy to specify the wrong generic. 
```cr
alias GuildCache = Cache(UInt64, Guild)
````

---
Todo: 
- [ ] Documentation

### CODE REVIEW

**Comment by z64** [comment]
The index caches are pointless and force you to use an array :disappointed: 

I'm not sure how to fix this right now.

**Comment by z64** [comment]
This works now:
```cr
require "./discordcr"
client = Discord::CachedClient.new ENV["TOKEN"]

client.configure_caches do
  guild_cache Discord::MemoryCache
  channel_cache Discord::MemoryCache
  role_cache Discord::MemoryCache
  user_cache Discord::MemoryCache
end
```
- I don't know how to capture `client` in `client.configure_caches` (if i name `client` something else, it breaks)
- I still don't know what to do about the index caches. Do we have another special `IndexCache` type?

z64:
The included `Enumerable#each` in `Cache` currently crashes the compiler if used :disappointed: 
```cr
require "./discordcr"
foo = Discord::CachedClient.new "token"

foo.configure_caches do
  all_caches Discord::MemoryCache
end

foo.guild_cache.each do |id, guild|
  p guild
end
```
```
BUG: trying to assign Discord::Guild <- UInt64
Error opening file 'crystal' with mode 'r': No such file or directory (Errno)
Failed to raise an exception: END_OF_STACK
[0x5602d693cff6] ???
[0x5602d613f299] __crystal_raise +153
[0x5602d6140f8c] ???
[0x5602d6151267] ???
[0x5602d614a906] ???
[0x5602d6149c2e] ???
[0x5602d61469c2] ???
[0x5602d6143866] main +422
[0x7fe22b958f4a] __libc_start_main +234
[0x5602d613c29a] _start +42
[0x0] ???
```
Tried a few other things, but couldn't figure out a viable workaround without hitting other bugs..

r00ster91:
Did you reported it on https://github.com/crystal-lang/crystal/issues ?

z64:
no, there are already open issues for it with no discussion of workarounds

z64:
I've rebased this for 0.25 and tidied up the commit history a little bit. 

---

Unfortunately `each` from my last example above still does not work.
```
Error in test/test.cr:23: BUG: already had enclosing call

  p guild
```
This looks like crystal issue #5299. Caches definitely need to be iterated on, so we need some kind of workaround  / different approach.

Here it is boiled down from our specific use case: https://carc.in/#/r/4bkg

This occurs "because" we instantiate the default with a `NullCache` but replace it with a `MemoryCache` later.

---

The only "fix" that I can see that works is telling the compiler with `as` what the *final* cache type is:
```cr
client.cache.as(MemoryCache(Int32, String)).each do |value|
  p value
end
```
https://carc.in/#/r/4bl3

Is an acceptable workaround to hack `.as` into our properties somehow? Other than this, I have no idea..

RX14:
@z64 you should ping the issue referencing that it's blocking this one.

z64:
Alright, I've rebased this for `Snowflake` compatibility (fdb5884) and "fixed" the `Enumerable` on `MemoryCache` (3dee944). I'll bump that issue. For now the workaround `.as` has to be used for `each`, etc.

Also during the last iteration, I believe we wanted to rename the base abstract `Cache` methods, but it wasn't clear exactly what was wanted;

- Did we want Hash-like `#[](k)`, `#[]=(k, v)`, `#delete(k)`
- Did we want `#get(k)`, `#put(k, v)`, `#delete(k)`
- Some other mix of those

z64:
Compiler bug fixed / worked around :smile:

src/discordcr/cache.cr
```diff
@@ -1,272 +1,66 @@
...
-    def cache(channel : Channel)
-      @channels[channel.id.to_u64] = channel
-    end
+    abstract def remove(key : K)
 
-    # Adds a specific *guild* to the cache.
-    def cache(guild : Guild)
-      @guilds[guild.id.to_u64] = guild
-    end
+    abstract def each(&block : Tuple(K, V) ->)
```
RX14: nah, just yield `K, V`. This is legacy from `Enumerable`

src/discordcr/cache.cr
```diff
@@ -1,272 +1,66 @@
...
+  end
 
-    # Adds a specific *member* to the cache, given the *guild_id* it is on.
-    def cache(member : GuildMember, guild_id : UInt64 | Snowflake)
-      guild_id = guild_id.to_u64
-      local_members = @members[guild_id] ||= Hash(UInt64, GuildMember).new
-      local_members[member.user.id.to_u64] = member
-    end
+  class MemoryCache(K, V) < Cache(K, V)
+    include Enumerable(Tuple(K, V))
```
RX14: Why?

z64: Ah yeah, I had wanted `cache.find` available etc. but of course you can just do `cache.each.find`. :+1:

RX14: If you want this, then put `include Enumerable({K, V})` in the abstract `Cache`, not each individual instance. Otherwise you can't use it.

z64: Okay, so I've done that. (`include Enumberable({K, V})` in `Cache`)

In your other review comment, if you mean to write:

```cr
abstract def each(&block : K, V ->)
```

And use just `: K, V ->` everywhere else, this works fine for `each`, however, there's some behaviour with `.find` I don't understand.

```cr
client.guild_cache.each do |id, guild|
  # Works, OK
end

# Doesn't compile: too many block arguments (given 2, expected maximum 1)
client.guild_cache.find do |id, guild|
  # ..
end

# Only the ID is yielded:
value = client.guild_cache.find do |id|
  # Works, OK
end

p typeof(value) # => Snowflake | Nil
```

This problem doesn't occur and all examples work if I use `&block : Tuple(K, V) ->` everywhere.

```cr
value = client.guild_cache.find do |id, guild|
  # OK
end

p typeof(value) # => (Tuple(Snowflake, Guild) | Nil)
```

RX14: Yeah, if you implement `Enumerable` you need to use `{K, V} ->`.

PR #137, opened 2018-05-15
Lazy fields

z64:
Discord's new "lazy guild" backend behavior has added `guild_id` to several objects, as well as a partial `GuildMember` object to messages.

Sample JSON with `message.guild_id` and `message.member`: https://gist.github.com/z64/7f6c07584c20868201f9f85dc2e74b57

This isn't officially documented yet; I've implemented these fields from a similar commit to `disco-py` from b1nzy. On that note I'm not 100% sure on the nilability of `PartialMember` fields, but I've got a test bot running on DAPI for a bit now with no issues reported.

This also adds caching of a messages author and member if possible, which we probably do want in order to cache invisible users that we may not have otherwise observed at all yet in any other cached event.


z64:
Rebased this for conflicts.

There is another G250K change: `Message#mentions` are now `Array(User)` that may contain a `PartialGuildMember` field. This is *only* filled in the `MESSAGE_CREATE` event. I've just added this to `User`:
```cr
 @mentions=
  [Discord::User(
    @avatar="208bcfc90f49915f25ec3808dcfa0f85",
    @bot=true,
    @discriminator="3729",
    @email=nil,
    @id=Discord::Snowflake(@value=213450769276338177_u64),
    @member=
     Discord::PartialGuildMember(
      @deaf=false,
      @joined_at=2016-10-24 17:01:33.196000000 +00:00,
      @mute=false,
      @nick=nil,
      @roles=
       [Discord::Snowflake(@value=226855600586358785_u64),
        Discord::Snowflake(@value=240172879361212416_u64),
        Discord::Snowflake(@value=244951251865960449_u64),
        Discord::Snowflake(@value=284014636582567936_u64)]),
    @mfa_enabled=nil,
    @username="Hornet",
    @verified=nil)],
```

PR #139, opened 2018-05-21
Update old API documentation links

r00ster91:
Right now the badge [docs](https://meew0.github.io/discordcr/doc/v0.1.0/) leads to docs from version 0.1.0


z64:
Duplicate of #77 , also read the conversation there. If we want to add a master docs badge that's up for discussion, but this should be a link to the latest tag.

r00ster91:
I changed to 0.3.0 now. But I noticed that https://meew0.github.io/discordcr/doc/v0.3.0/ doesn't even exists.
So I think we should just link to master so we don't need to update every time again. Or someone builds the 0.3.0 docs.

z64:
yes @meew0 can fix the docs build

meew0:
The fact that tags don't cause doc builds should have been fixed shortly after the last release. I'll update the badge in the next release

PR #114, opened 2018-01-21
Switch to new Time Location format

greenbigfrog:
Latest (master) crystal introduces a new format for UTC time. Hold on merging till new crystal version releases.

PR #141, opened 2018-06-17
Changes for Crystal 0.25

z64:
Surpasses #114 (that change is actually wrong)

The location of timestamps is now parsed out of the format, specifying `Kind` (or its successor, `Location`) is not necessary.

Additionally, the formats between `TimestampConverter` and `EmbedTimestampConverter` were so similar and compatible, that I've removed it and use `TimestampConverter` everywhere.

I also do this because `Time::Format::ISO_8601_DATE`/`DATE_TIME` no longer defines `.from_json`/`.to_json`. 

I've tested this for a while on a live connection as well as some chat logs I had laying around on disk from previous issues with parsing timestamps with no issues.

### DIFF
--- a/spec/discordcr_spec.cr
+++ b/spec/discordcr_spec.cr
@@ -61,7 +61,7 @@ describe Discord do
     end
 
     it "serializes" do
-      json = %({"data":"2017-11-16T13:09:18.291+00:00"})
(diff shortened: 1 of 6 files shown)


z64:
Thanks! :)

PR #130, opened 2018-04-26
Implement Snowflake type

z64:
Closes #106 

~~Still WIP, as the specs are failing as somehow the nanosecond is lost between `Time.epoch_ms` and `Time#epoch_ms`. I haven't dug into it to find out why / how to fix it: https://carc.in/#/r/3ych~~

Enough operators should be delegated to allow snowflakes to be sorted in a collection.

PR #142, opened 2018-06-23
Fix Snowflake interpolation, mention_roles type

z64:
Updated my main bot to v0.4 and got some leftover things from #130 (Sorry! :disappointed:)

Right now `Snowflake` doesn't interpolate into HTTP routes well, so using `Snowflakes` in `REST` fails. I've fixed this by `def to_s(io : IO)`, writing the internal `@value`.
```cr
# current:
s = Discord::Snowflake.new(123_u64)
puts "/some/#{s}/route" # => "/some/Discord::Snowflake(@value=123_u64)/route"

# fixed:
s = Discord::Snowflake.new(123_u64)
puts "/some/#{s}/route" # => "/some/123/route"
```
Workaround is to use, i.e., `message.channel_id.to_u64` (or patch this yourself)

Secondly, looks like I just missed the `?` in subbing out an `Array(UInt64)?` type restriction. I reviewed #130 again and it looks like this is the only one I missed.

### DIFF
--- a/src/discordcr/mappings/gateway.cr
+++ b/src/discordcr/mappings/gateway.cr
@@ -309,7 +309,7 @@ module Discord
         tts: Bool?,
         mention_everyone: Bool?,
         mentions: Array(User)?,
-        mention_roles: Array(Snowflake),
+        mention_roles: Array(Snowflake)?,
         attachments: Array(Attachment)?,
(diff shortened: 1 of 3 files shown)


vizv:
I had the same issue, and this pr fixes it

PR #143, opened 2018-06-23
Fix examples

r00ster91:
Since the 0.4.0 update things like `client.create_message(payload.channel_id, "Pong!")` don't work anymore. This fixes and improves the examples a bit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ dependencies:
 ## Usage
 
 Example bots can be found
-[here](https://github.com/meew0/discordcr/blob/master/examples/ping.cr). More
+[here](https://github.com/meew0/discordcr/blob/master/examples/). More
 examples will come in the future.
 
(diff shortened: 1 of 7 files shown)


r00ster91:
Oh.. Then I will remove all the `.value`s but can we maybe still merge the improvements then? They are not just cosmetic changes. It also fixes wrong substraction in [examples/ping_with_response_time.cr](https://github.com/r00ster91/discordcr/blob/21ea6bcf82107f21e05cb38068127b391989519e/examples/ping_with_response_time.cr#L13) (it gives a negative value) and uses chars where chars can be used.

z64:
If the time is negative in that example, then your PC clock is likely out of sync. https://carc.in/#/r/4diw

I agree with the changes to `README.md`. The rest I don't have an opinion on.

PR #144, opened 2018-06-30
Decode close message

z64:
Adds decoding of the binary string from the close frame per [RFC 6455 §5.5.1](https://tools.ietf.org/html/rfc6455#section-5.5.1)

Before:
```
W, [2018-06-30 09:48:29 -04:00 #23180]  WARN -- discordcr: [Client] Closed with: �
```
After:
```
W, [2018-06-30 09:48:07 -04:00 #23094]  WARN -- discordcr: [Client] Closed with: 1002
```

### DIFF
--- a/src/discordcr/client.cr
+++ b/src/discordcr/client.cr
@@ -164,7 +164,14 @@ module Discord
 
     private def on_close(message : String)
       # TODO: make more sophisticated
-      @logger.warn "[#{@client_name}] Closed with: " + message
+      code = nil
+      reason = nil
+      unless message.empty?
+        io = IO::Memory.new(message)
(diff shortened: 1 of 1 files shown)


RX14:
Thats a stdlib bug.

z64:
Alright. If I'm reading websocket's implementation right, decoding the close code just isn't implemented. They just stringify the buffer and pass it off to the user's handler. Is that the issue?

RX14:
Yes. Standard says MUST and we just ignore it and treat it as a standard string buffer,

RX14:
Please make an issue for this on the crystal repo.

meew0:
https://github.com/crystal-lang/crystal/issues/6469

PR #152, opened 2018-07-12
Fix return type for Cache#resolve_dm_channel

greenbigfrog:
Introduced with change to Discord::Snowflake

(Actually I'm not sure whether it should be returning a UInt64 or Discord::Snowflake, but this is what I (was) expecting)


RX14:
pretty sure this should be returning snowflakes.

z64:
@RX14 the current iteration of the cache stores IDs as UInt64;
```cr
@dm_channels = Hash(UInt64, UInt64).new
```
 We talked about it when I was writing the snowflake PR and decided to leave Cache alone for now. The cache in #136 uses snowflakes

RX14:
ok

PR #154, opened 2018-07-18
Implement Discord::Paginator

### LINKED ISSUE #140: Iterators for paginated endpoints
It would be great if we provided iterators for paginated endpoints, namely for traversing channel message history to backfill logs or guild members for stateless apps over REST.

What would be the preferred API for this? For example, should we provide a different method/overload, something like:
```cr
def get_channel_messages(channel_id, start_message_id = 1_u64, direction = Direction::Down, limit = nil)
```
and similar for other endpoints?

z64:
Implements a paginator utility class for making requests across paginated endpoints seamlessly.

This is useful for:
-  fetching large volumes of channel history for backfilling a chat log database 
- a means of examining all of the guilds a client is in if they're in more than 100
- listing all of the members of a guild if there are more than 1000

---

Example:

```cr
client.page_channel_messages(381889335278043137, limit: 20, page_size: 5).each do |message|
  p({message.id.to_u64, message.author.username, message.content})
end
```
([Results](https://gist.github.com/z64/74413410bc11623d857e8aee191a5abc))

Closes #140

### DIFF
--- a/src/discordcr/rest.cr
+++ b/src/discordcr/rest.cr
@@ -225,7 +225,7 @@ module Discord
     # Message History" permission.
     #
     # [API docs for this method](https://discordapp.com/developers/docs/resources/channel#get-channel-messages)
-    def get_channel_messages(channel_id : UInt64 | Snowflake, limit : UInt8 = 50, before : UInt64 | Snowflake | Nil = nil, after : UInt64 | Snowflake | Nil = nil, around : UInt64 | Snowflake | Nil = nil)
+    def get_channel_messages(channel_id : UInt64 | Snowflake, limit : Int32 = 50, before : UInt64 | Snowflake | Nil = nil, after : UInt64 | Snowflake | Nil = nil, around : UInt64 | Snowflake | Nil = nil)
       path = "/channels/#{channel_id}/messages?limit=#{limit}"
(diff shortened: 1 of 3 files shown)


src/discordcr/rest.cr
```diff
@@ -243,6 +243,27 @@ module Discord
       Array(Message).from_json(response.body)
     end
 
+    # Returns an `Paginator` over a channel's message history. Requires the
```
meew0: A bit nitpicky but this should be ``a `Paginator` ``

z64:
@meew0 comments from @RX14 in the past suggested it wasn't worth using types other than Int32 in REST. Though, this was before automatic casting of literals was added. I'm fine with changing it back, I think it just got in my way while writing this.

RX14:
There's no point in using the smallest number literal type possible unless the range is *exactly* the same. Int32 and Int64 are really the only types that should be used in high level code, the rest are just there for completeness and for low-level programming. If say Int16's broke I don't think anyone would notice for quite some time.

meew0:
That makes sense. @z64 can you fix the typo and rebase?

z64:
Sure thing.

z64:
Updated

PR #155, opened 2018-07-23
Add utility methods for CDN resources

z64:
Should `Discord::User#avatar_url`, etc be created that wraps these functions, or perhaps `CDN.user_avatar(user : User)`?

### DIFF
--- a/spec/cdn_spec.cr
+++ b/spec/cdn_spec.cr
@@ -58,3 +58,151 @@ describe Discord::CDN do
     end
   end
 end
+
+describe Discord::User do
+  user_with_default_avatar = Discord::User.from_json <<-JSON
+  {
+    "id": "1",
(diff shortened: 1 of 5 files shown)


RX14:
the `to_s` without io on those enums should be downcase too.

z64:
@RX14 Don't think so.. :confused: https://carc.in/#/r/4l0m

RX14:
guess you're doing to have to do it manually with `case` then

z64:
Sure. I was just thinking as having used discordcr a lot, I don't expect the structs to provide anything Discord's API doesn't directly provide, and a CDN URL isn't strictly that. Just wanted to check before setting a precedent for this kind of abstraction on our structs.

I've added relevant CDN methods to structs.

Issue #90, opened 2017-08-31
Specs


z64:
It would be good to set up a policy on specs (`CONTRIBUTING.md`) and/or adopt a spec suite. I'm partial to trying out `minitest.cr`, but I'm not fully aware of the other options available, as well as specs practices in Crystal (if they differ in any way from langs like ruby w/ rspec). The crystal stdlib spec API seems a bit limited.

z64:
In other libraries I've written for REST APIs, I've created a namespace for functions that generate `HTTP::Request`, for example:
```cr
# request.cr
module Request
  extend self

  def get_resource(param)
    params = HTTP::Params.build do |form|
      form.add("param", param)
    end

    HTTP::Request.new("GET", "/resource?#{params}")
  end
end

# rest.cr
module REST
  @http_client = HTTP::Client.new(API_HOST, tls: true)

  def request(request : HTTP::Request)
    @http_client.exec(request)
  end

  def get_resource(param)
    response = request(Request.get_resource(param))
    Resource.from_json(response.body)
  end
end
```

Allowing me to create a spec:
```cr
it "forms GET /resource" do
  request = Request.get_resource("foo")
  request.method.should eq "GET"
  request.path.should eq "/resource"
  request.query.should eq "param=foo"  
end
```

I could write all of these `Request` functions and specs per the API documentation, and then write my `REST` binding knowing that I was always executing a correctly formed request.

Would there be value in writing a similar request-building function layer for discordcr?

RX14:
Is there still no mock API server, wasn't there a project to make one?

z64:
A few - meew0 was going to make one ([link](https://github.com/meew0/l-)), as well as b1nzy at one point I think. Maybe others. None to my knowledge are functional for any serious use

PR #158, opened 2018-08-07
[WIP] Add mock HTTP server

z64:
This PR adds a simple mock HTTP server that runs alongside our specs so that we can test runtime behavior of our REST functions! :smile: :heart: 

---

## How it works:

The server exposes a simple API of `POST /_endpoints` to "prepare" and endpoint with a response with an object like the following:
```json
{
  "method": "GET",
  "path": "/ping",
  "status_code": 200,
  "headers": {
    "Content-Type": "text/plain"
  },
  "body": "pong"
}
```
You can then `GET /ping` and receive that object as you specified, with the given headers and status code applied. The `Client` is pointed at our mock server with `-Dmock_server` as a compile-time change.

This lets us write specs using a class-level function that prepares an endpoint:
```cr
describe Discord::REST do
  it "#create_message" do
    # Instance the expected object
    expected = Discord::Message.from_json(message_json)

    # Tell the server what we expect to receive back on the next call to this endpoint
    Discord::MockServer.prepare_endpoint("POST", "/channels/1/messages", 200,
      {"Content-Type" => "application/json"}, expected.to_json)

    # Spec it!
    client.create_message(1, expected.content).should eq expected
  end
end
```

CI is accomplished by `spec_runner.sh` that starts the server in the background, waits until its prepared `/ping` endpoint replies, runs specs, then kills the process.

---

## This PR

Mostly I am looking for feedback on the design of the Mock server and its integration into the library and specs. There are still some TODO left around that I'll touch up tomorrow probably.

There are *a lot* of REST functions, so if possible, I would like this PR to merge the server and an example spec, and more REST coverage can come in later PRs to keep this manageable.

My main concerns right now:

- Should the server be under `src/` (maybe `src/mock_server`) for users to use in their own bots? Or should it be kept as a lib developer tool?

- I'm unsure that logging is needed server side but I added it anyways. It just logs to INFO right now.

- `spec_runner.sh`. I don't pretend to know bash, this was certainly hacked together, but it seems to work.

- There's going to be a LOT of JSON in these specs. Should we add a `spec/json/` dir and a JSON loader helper method to keep specs easier on the eyes? Or do we care?


RX14:
we don't need a mock HTTP server, just to use something like webmock

z64:
Hm, I think I'd rather still stick with our own solution. Eventually we'll want to mock the gateway too which webmock doesn't cover, and this pr would give us a pretty easy path to. Or we'd have to find some other shard.
(Also, a webmock spec is failing for me on 0.25, I'm not sure why or if it's relevant at the moment)

If we want to go with webmock for now, happy to throw that together; from #90 and other convos something custom seemed best in lieu of there not being any well-developed Discord mock services.

spec/mock_server.cr
```diff
@@ -0,0 +1,105 @@
...
+    request, response = context.request, context.response
+    route = {request.method, request.resource}
+    handled = false
+
+    begin
+      case route
+      when {"GET", "/_endpoints"}
+        respond_json(@endpoints.values.to_json, 200, response)
+        handled = true
+      when {"POST", "/_endpoints"}
```
RX14: why?

spec/mock_server.cr
```diff
@@ -0,0 +1,105 @@
...
+      response.headers.add(key, value)
+    end
+    response.status_code = endpoint.status_code
+    response.content_length = endpoint.body.bytesize
+    response.print(endpoint.body)
+  end
+end
+
+# TODO: make nicer
+if ARGV[0]? == "run"
```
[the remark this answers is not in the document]
z64: I spec the MockServer. it would run when I ran its own specs

spec/mock_server.cr
z64: Sorry, could you be more specific?

PR #160, opened 2018-08-16
Add query params to #create_guild_ban, Add #get_guild_ban

z64:


src/discordcr/rest.cr
```diff
@@ -1030,15 +1030,38 @@ module Discord
       Array(GuildBan).from_json(response.body)
     end
 
+    # Returns information about a banned user in a guild. Require the "Ban Members"
```
r00ster91: Typo: Require

r00ster91: I meant this should be "Requires".

PR #161, opened 2018-08-18
Support zlib-stream gateway compression

z64:
This PR adds support for Discord's `zlib-stream` compression transport mode. (Ref #10 )

## Background

Initially I had a ton of difficulty with this because the stdlib's `Zlib::Reader` basically isn't meant to do this; It is really just meant to do one-shot inflation.

The first issue is that when you `Reader.new(io)`, it immediately tries to consume the zlib header from that IO. This isn't helpful for us because we basically need a single instance of a `Reader` to keep around forever, and having `@reader : Reader?` is annoying.

~~To fix this, I've subclassed `Zlib::Reader` into `Zlib::StreamReader` that lets us control when we read the header - *only* on the very first binary frame we receive, letting us keep a single instance around. I've made an instance method that calls `Zlib::Reader.read_header` as it is `protected`.~~

\* - **See updated diff; we now lazily initialize the stdlib `Zlib::Reader`**

Next, if you just pass `Object.from_json(zlib_reader)`, this will deadlock the program. The *entire* program as far as I can tell. Example: https://github.com/z64/zlib-freeze

\* - **Upstream issue:** https://github.com/crystal-lang/crystal/issues/6575

This doesn't happen if you `reader.read(slice)`. So I've created a single 10MB  `Bytes` buffer (referencing b1nzy's [disco implementation](https://github.com/b1naryth1ef/disco/blob/master/disco/gateway/client.py#L178)) specifically to throw the inflated JSON into, which we can then wrap in an IO and pass to our objects.

Lastly, to avoid a mess of a single binary handler, I've split our `on_binary` into two methods: `on_compressed` for the basic "large" compression mode, and `on_compressed_stream` for `zlib-stream` frames that both register their own `@websocket.on_binary` handlers respectively.

## Interface

`compressed` in `Client#initialize` is now an enum instead of bool:
```cr
client = Discord::Client.new(token: "token", compress: :none)
client = Discord::Client.new(token: "token", compress: :large)
client = Discord::Client.new(token: "token", compress: :stream)
```

`CompressMode::Stream` is now the default, as it has the most bandwidth savings, and per rumor, might be required for the next major gateway version.

The appropriate callbacks and IDENTIFY payloads are configured by checking `@compress.large?` and `@compress.stream?`. 

I've had a bot running with this `zlib-stream` code for about 12 hours without issue, but please feel free to test this on your own.

### DIFF
--- a/src/discordcr/websocket.cr
+++ b/src/discordcr/websocket.cr
@@ -56,7 +56,8 @@ module Discord
     @zlib_reader : Zlib::Reader?
     @buffer : Bytes
 
-    def initialize(@host : String, @path : String, @port : Int32, @tls : Bool, @logger : Logger)
+    def initialize(@host : String, @path : String, @port : Int32, @tls : Bool, @logger : Logger,
+                   @zlib_buffer_size : Int32 = 10 * 1024 * 1024)
       @websocket = HTTP::WebSocket.new(
         host: @host,
         path: @path,
@@ -64,8 +65,8 @@ module Discord
         tls: @tls
       )
 
-      # 10MB buffer for zlib-stream
-      @buffer_memory = Bytes.new(10 * 1024 * 1024)
+      # Buffer for zlib-stream
+      @buffer_memory = Bytes.empty
       @buffer = @buffer_memory[0, 0]
       @zlib_io = IO::Memory.new
       @zlib_reader = nil
@@ -83,6 +84,7 @@ module Discord
     end
 
     def on_compressed_stream(&handler : Packet ->)
+      @buffer_memory = Bytes.new(@zlib_buffer_size)
       @websocket.on_binary do |binary|
         @zlib_io.write binary
         next if binary.size < 4 || binary[binary.size - 4, 4] != ZLIB_SUFFIX
--- a/src/discordcr/websocket.cr
+++ b/src/discordcr/websocket.cr
@@ -51,6 +51,11 @@ module Discord
       end
     end
 
+    ZLIB_SUFFIX = Bytes[0x0, 0x0, 0xFF, 0xFF]
+
+    @zlib_reader : Zlib::Reader?
+    @buffer : Bytes
+
     def initialize(@host : String, @path : String, @port : Int32, @tls : Bool, @logger : Logger)
       @websocket = HTTP::WebSocket.new(
         host: @host,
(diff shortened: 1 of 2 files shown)


PixeLInc:
Finally. Now I don't have to hear complaints about this all the time ;)

z64:
Just squashed my fixups ~~so meew0 doesn't merge them~~. Thanks for reviewing! :smile:

src/discordcr/websocket.cr
```diff
@@ -51,16 +51,27 @@ module Discord
...
     def initialize(@host : String, @path : String, @port : Int32, @tls : Bool, @logger : Logger)
       @websocket = HTTP::WebSocket.new(
         host: @host,
         path: @path,
         port: @port,
         tls: @tls
       )
+
+      # 10MB buffer for zlib-stream
+      @buffer_memory = Bytes.new(10 * 1024 * 1024)
```
meew0: It seems unnecessary to allocate this unless we're actually using stream compression, so could you maybe change it to only allocate the buffer in that case? 10MB isn't much in the large scale of things but this seems like an easy way to save some memory.

src/discordcr/websocket.cr
```diff
@@ -71,6 +82,25 @@ module Discord
...
 
+    def on_compressed_stream(&handler : Packet ->)
+      @websocket.on_binary do |binary|
+        @zlib_io.write binary
+        next if binary.size < 4 || binary[binary.size - 4, 4] != ZLIB_SUFFIX
+        @zlib_io.rewind
+
+        zlib_reader = (@zlib_reader ||= Zlib::Reader.new(@zlib_io))
+
+        read_size = zlib_reader.read(@buffer_memory)
```
meew0: I think this `read` call should be looped so the whole packet is read, in case a packet >10MB is received. (Seems unlikely but who knows what Discord might do) This would also allow the buffer size to be configurable in case people want to save some more memory

src/discordcr/websocket.cr
z64: Sure, I can make that happen when the callback is registered 👍

src/discordcr/websocket.cr
z64: I'll look into this; and yeah I just pulled 10MB from b1nzy's disco lib and figured it was a safe bet.

Do you actually want me to expose some interface to make this configurable? If so I'm not sure off the top of my head the best place to do this, other than to pass it down from the client initializer..

meew0: Yeah I think passing it down from the client initializer is the best bet, I agree it's a bit clunky but I don't see a better way either

PR #168, opened 2018-11-18
Fix link to docs in readme (wrong version)

ChallahuAkbar:
Whoopsies...

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ puts user.username
 ```
 
 Apart from this, API documentation is also available, at
-https://meew0.github.io/discordcr/doc/v0.1.0/.
+https://meew0.github.io/discordcr/doc/v0.4.0/.
 
 ## Contributing
 


README.md
```diff
@@ -72,7 +72,7 @@ puts user.username
 ```
 
 Apart from this, API documentation is also available, at
-https://meew0.github.io/discordcr/doc/v0.1.0/.
+https://meew0.github.io/discordcr/doc/v0.4.0/.
```
r00ster91:
```suggestion
https://meew0.github.io/discordcr/doc/master/.
```
This should be just master so it doesn't have to be updated every version. Also, pretty much everyone uses master.

z64: No, we can provide links to both if we really want to. There should always be a link to the latest tag.

greenbigfrog: How many breaking changes can we really expect to happen in the near future.
IMO linking to master is totally fine in this case.

z64: You can expect everything to break because we're `v0.x`.

Linking to master is fine, but it should not replace the existing latest tag link.

kandayo: Since the latest version is broken and no one is using it I don't see the point in linking to its docs.

z64: No, *the point* is to fix our release cycle so that the latest version *can* be used.

Our weak release cycle isn't an excuse to change versioning practices / semantics - that tags should be relied on, and `master` or otherwise for bleeding edge changes / fixes.

(Even still, v0.4 is fine with a single monkey patch you can apply in about 5 lines.)

PR #169, opened 2018-11-25
Fix @client_id must be UInt64 not Snowflake

kandayo:
```
in lib/discordcr/src/discordcr/client.cr:118: 
instance variable '@client_id' of Discord::Client must be (UInt64 | Nil), not Discord::Snowflake
  @client_id ||= get_oauth2_application.id
```

### DIFF
--- a/src/discordcr/client.cr
+++ b/src/discordcr/client.cr
@@ -84,7 +84,7 @@ module Discord
     # The *properties* define what values are sent to Discord as analytics
     # properties. It's not recommended to change these from the default values,
     # but if you desire to do so, you can.
-    def initialize(@token : String, @client_id : UInt64? = nil,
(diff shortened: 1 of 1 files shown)


src/discordcr/client.cr
```diff
@@ -84,7 +84,7 @@ module Discord
     # The *properties* define what values are sent to Discord as analytics
     # properties. It's not recommended to change these from the default values,
     # but if you desire to do so, you can.
-    def initialize(@token : String, @client_id : UInt64? = nil,
+    def initialize(@token : String, @client_id : (UInt64 | Snowflake)? = nil,
```
z64: To be consistent with other type union styles, we've been writing out explicitly `UInt64 | Snowflake | Nil` if there's more than one other type.

PR #170, opened 2018-11-27
Snowflake#>>

kandayo:
It would be useful since it's used by the sharding formula.


z64:
I don't think this is necessary. If we want the shard formula, we should implement it.

kandayo:
You have a point. 
But it feels weird to cast guild_id to_u64 just to use it. Since the formula is defined on the Discord API docs, Snowflakes *can* be right-shifted.

PR #174, opened 2018-12-17
Fix edit_channel_permissions' docs and add missing permissions

r00ster91:
This fixes a wrong permission in [`edit_channel_permissions`](https://meew0.github.io/discordcr/doc/master/Discord/REST.html#edit_channel_permissions(channel_id:UInt64|Snowflake,overwrite_id:UInt64|Snowflake,type:String,allow:Permissions,deny:Permissions)-instance-method)' docs and removes the comment in `Permissions.new` (see https://github.com/meew0/discordcr/pull/174#issuecomment-447901908).
It also adds 2 missing permissions to the `Permissions` enum.

### DIFF
--- a/src/discordcr/mappings/permissions.cr
+++ b/src/discordcr/mappings/permissions.cr
@@ -8,6 +8,8 @@ module Discord
     ManageChannels      = 1 << 4
     ManageGuild         = 1 << 5
     AddReactions        = 1 << 6
+    ViewAuditLog        = 1 << 7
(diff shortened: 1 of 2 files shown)


z64:
I don't think we should do this. There are gaps in the permissions flags, and if Discord were to start making use of them in the API, this makes every object that uses `Permissions` fail to instantiate.

The strictness is good when you control the API, but in this case we don't, and being inflexible here is just inconvenient for no real benefit.

PR #175, opened 2018-12-21
Update Emoji endpoints

z64:
- Add `REST#get_guild_emoji`
- Add `REST#delete_guild_emoji`
- Rename `REST#get_guild_emojis` to `REST#list_guild_emojis`
- Fix inconsistent rate limit keys in `list_guild_emoji`, `modify_guild_emoji`, and `create_guild_emoji`

### DIFF
--- a/src/discordcr/rest.cr
+++ b/src/discordcr/rest.cr
@@ -701,7 +701,7 @@ module Discord
     # [API docs for this method](https://discordapp.com/developers/docs/resources/emoji#list-guild-emojis)
     def list_guild_emojis(guild_id : UInt64 | Snowflake)
       response = request(
-        :guild_gid_emojis,
+        :guilds_gid_emojis,
         guild_id,
         "GET",
         "/guilds/#{guild_id}/emojis",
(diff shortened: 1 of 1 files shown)


src/discordcr/rest.cr
```diff
@@ -765,6 +765,20 @@ module Discord
       Emoji.from_json(response.body)
     end
 
+    # Deletes a guild emoji. Requires the "Manage Emojis" permission.
+    #
+    # [API docs for thsi method](https://discordapp.com/developers/docs/resources/emoji#delete-guild-emoji)
+    def delete_guild_emoji(guild_id : UInt64 | Snowflake, emoji_id : UInt64 | Snowflake)
```
RX14: ` : Nil`?

z64: Sure. I would rather do that in another PR to decorate every method, this would be the only place in `rest.cr` that we have a return type.

src/discordcr/rest.cr
```diff
@@ -749,6 +765,20 @@ module Discord
       Emoji.from_json(response.body)
     end
 
+    # Deletes a guild emoji. Requires the "Manage Emojis" permission.
+    #
+    # [API docs for thsi method](https://discordapp.com/developers/docs/resources/emoji#delete-guild-emoji)
```
r00ster91:
```suggestion
    # [API docs for this method](https://discordapp.com/developers/docs/resources/emoji#delete-guild-emoji)
```

PR #179, opened 2019-02-02
Add spoiler argument to Client#upload_file

r00ster91:
This adds a `spoiler` argument to [`Client#upload_file`](https://meew0.github.io/discordcr/doc/master/Discord/REST.html#upload_file%28channel_id%3AUInt64%7CSnowflake%2Ccontent%3AString%3F%2Cfile%3AIO%2Cfilename%3AString%3F%3Dnil%29-instance-method) which marks the image as a spoiler and is just the same as: 
image

### DIFF
--- a/src/discordcr/rest.cr
+++ b/src/discordcr/rest.cr
@@ -404,7 +404,7 @@ module Discord
         end
       end
 
-      if spoiler
+      if spoiler && !filename.starts_with?("SPOILER_")
         filename = "SPOILER_" + filename
       end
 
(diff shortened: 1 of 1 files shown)


z64:
Provisioning Crystal on Travis has been fixed. Can you please re-trigger CI so it can complete?

src/discordcr/rest.cr
```diff
@@ -404,6 +404,10 @@ module Discord
         end
       end
 
+      if spoiler
```
z64: Maybe we should avoid prepending this again if it is already provided by the user or `File`.
```suggestion
      if spoiler && !filename.starts_with?("SPOILER_")
```

PR #180, opened 2019-02-02
Remove Client#accept_invite and use bulk-delete endpoint instead of bulk_delete

r00ster91:
The current endpoint that is used for bulk-deleting is deprecated:
https://discordapp.com/developers/docs/resources/channel#bulk-delete-messages-deprecated
Now the new one is used.
This also removes [`Client#accept_invite`](https://meew0.github.io/discordcr/doc/master/Discord/REST.html#accept_invite%28code%3AString%29-instance-method) which has been deprecated as well and is already discontinued (no longer works):
https://discordapp.com/developers/docs/change-log#deprecation-accept-invite-endpoint

### DIFF
--- a/src/discordcr/client.cr
+++ b/src/discordcr/client.cr
@@ -766,7 +766,7 @@ module Discord
     event message_delete, Gateway::MessageDeletePayload
 
     # Called when multiple messages are deleted at once, due to a bot using the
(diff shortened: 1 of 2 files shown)


z64:
Provisioning Crystal on Travis has been fixed. Can you please re-trigger CI so it can complete?

PR #185, opened 2019-04-18
Add REST#modify_guild_role_positions method

deingithub:
Allows users to modify the order of roles: https://discordapp.com/developers/docs/resources/guild#modify-guild-role-positions


deingithub:
Not sure whether I can explain that, tbh.

z64:
@deingithub It's just a change in 0.28 (that was just released) that breaks one of our specs. NBD.

PR #188, opened 2019-05-27
Add GamePlaying::Type

r00ster91:
This adds the `Type` enum to `GamePlaying` and makes it the required type for the `type` argument. It also removes the `String` type from `type` since that's a legacy type for that field.

### DIFF
--- a/src/discordcr/mappings/guild.cr
+++ b/src/discordcr/mappings/guild.cr
@@ -245,12 +245,21 @@ module Discord
   end
 
   struct GamePlaying
-    def initialize(@name = nil, @type = nil, @url = nil)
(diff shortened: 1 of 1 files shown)


src/discordcr/mappings/guild.cr
```diff
@@ -245,12 +245,21 @@ module Discord
...
+    def initialize(@name = nil, @type : Type? = nil, @url = nil)
+    end
+
+    # All the activities for the status of the current user.
+    enum Type
+      Playing = 0
+      # This activity only has effect when `url` is a Twitch URL.
+      Streaming = 1
+      Listening = 2
+      Watching = 3
```
GeopJr:
```suggestion
      Watching = 3
      Custom = 4
```

PR #190, opened 2019-05-30
Implement the Create Guild endpoint

ghost:
This implements the missing Create Guild endpoint and fixes `Discord::REST#delete_guild` erroring because of the `Guild.from_json(response.body)`. The endpoint actually returns `204 No Content` on success, so nothing in discordcr.

### DIFF
--- a/src/discordcr/mappings/guild.cr
+++ b/src/discordcr/mappings/guild.cr
@@ -237,13 +237,13 @@ module Discord
     end
   end
 
-  struct PartialRole
+  struct GuildCreateRole
     include JSON::Serializable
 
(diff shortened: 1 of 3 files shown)


src/discordcr/mappings/guild.cr
```diff
@@ -237,6 +237,23 @@ module Discord
     end
   end
 
+  struct PartialRole
```
r00ster91: I'm not sure this is a good name because every field from `Role` is here too but I think an additional type is required because here all the instance variable types are nilable, in `Role` they are not.
Something like `GuildCreationRole` might be too long.

PR #200, opened 2020-01-07
Voice Update (v4)

swarley:
This PR fixes the voice UDP connection process, and adds support for `xsalsa20_poly1305_suffix` and `xsalsa20_poly1305_lite`.

This has been tested to work locally with the `voice_send` example on Crystal 0.32.1

I'm not well versed in idiomatic crystal conventions so I welcome any reviews or suggestions.

## Added
`xsalsa20_poly1305_suffix` and `xsalsa20_poly1305_lite` support
`ENCRYPTION_MODES` constant
`lite_nonce` instance variable in `VoiceUDP` for tracking the incremental nonce.

## Changed
Voice websocket v4 is now requested.
Heartbeats now have a nonce in the data field.
Remove heartbeat interval from `READY` payload, as it is listed as an erroneous field
Heartbeat interval is now a `Float32`, and taken from the `HELLO` payload exclusively.

## Fixed
UDP connection now gets the server IP from the proper payload.

## Removed
`ENCRYPTED_MODE` constant

### DIFF
--- a/src/discordcr/voice.cr
+++ b/src/discordcr/voice.cr
@@ -100,18 +100,10 @@ module Discord
       @udp.send_audio(buf, @sequence, @time)
     end
 
-    private def wrapped_add(num : Number, increment : Number, max : Number)
-      if (num + increment) > max
-        0
-      else
-        num + increment
-      end
-    end
-
     # Increment sequence and time
     private def increment_packet_metadata
-      @sequence = wrapped_add(@sequence, 1, 0xff_ff).as(UInt16)
-      @time = wrapped_add(@time, 960, 0xff_ff_ff_ff).as(UInt32)
+      @sequence &+= 1
+      @time &+= 960
     end
 
     private def heartbeat_loop
@@ -289,11 +281,7 @@ module Discord
         nonce = Bytes.new(4)
         IO::ByteFormat::BigEndian.encode(@lite_nonce, nonce)
 
-        @lite_nonce = if @lite_nonce >= 0xff_ff_ff_ff
-                        0_u32
-                      else
-                        @lite_nonce + 1
-                      end
+        @lite_nonce &+= 1
       else
         raise "Cannot create a nonce for unsupported audio mode `#{@mode}'"
       end
--- a/src/discordcr/voice.cr
+++ b/src/discordcr/voice.cr
@@ -269,7 +269,7 @@ module Discord
       bytes
     end
 
-    def create_nonce(header : Bytes)
+    private def create_nonce(header : Bytes)
       nonce = nil
       case @mode
       when "xsalsa20_poly1305"
--- a/src/discordcr/voice.cr
+++ b/src/discordcr/voice.cr
@@ -9,8 +9,9 @@ module Discord
   class VoiceClient
     UDP_PROTOCOL = "udp"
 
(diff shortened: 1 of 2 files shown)


src/discordcr/voice.cr
```diff
@@ -9,9 +9,12 @@ module Discord
   class VoiceClient
     UDP_PROTOCOL = "udp"
 
-    # The mode that tells Discord we want to send encrypted audio
+    # DEPRECATED: Discord now supports multiple encryption modes
     ENCRYPTED_MODE = "xsalsa20_poly1305"
```
RX14: Why not remove this?

swarley: Was leaving it in because I wasn't sure if this was considered to be part of the public API or if it needed to be phased out over a release cycle, etc.

I can remove it if it's considered to be for internal use only, which I would imagine it is.

src/discordcr/voice.cr
```diff
@@ -251,11 +269,30 @@ module Discord
...
+        header.copy_to(nonce)
+      when "xsalsa20_poly1305_suffix"
+        nonce = Random::Secure.random_bytes(24)
+      when "xsalsa20_poly1305_lite"
+        nonce = Bytes.new(4)
+        IO::ByteFormat::BigEndian.encode(@lite_nonce, nonce)
+
+        @lite_nonce &+= 1
+      else
+        raise "Cannot create a nonce for unsupported audio mode `#{@mode}'"
```
z64:
```suggestion
        raise "Cannot create a nonce for unsupported audio mode #{@mode.inspect}"
```

src/discordcr/voice.cr
z64: You can remove it.

src/discordcr/voice.cr
```diff
@@ -146,16 +146,22 @@ module Discord
...
-      # We get a new heartbeat interval here that replaces the old one
-      @heartbeat_interval = payload.heartbeat_interval
-      udp_connect(payload.port.to_u32, payload.ssrc.to_u32)
+      supported_modes = ENCRYPTION_MODES & payload.modes
+
+      target_mode = if supported_modes.empty?
+                      raise "No supported modes found. Available: #{payload.modes}"
+                    else
+                      supported_modes.first
+                    end
```
z64:
```cr
private def handle_ready(payload : VWS::ReadyPayload)
  supported_modes = ENCRYPTION_MODES & payload.modes
  target_mode = supported_modes.first? || raise "No supported modes found in #{payload.modes}"
  udp_connect(payload.ip, payload.port.to_u32, payload.ssrc.to_u32, target_mode)
end
```

src/discordcr/voice.cr
```diff
@@ -251,11 +269,30 @@ module Discord
...
+        nonce = Random::Secure.random_bytes(24)
+      when "xsalsa20_poly1305_lite"
+        nonce = Bytes.new(4)
+        IO::ByteFormat::BigEndian.encode(@lite_nonce, nonce)
+
+        @lite_nonce &+= 1
+      else
+        raise "Cannot create a nonce for unsupported audio mode `#{@mode}'"
+      end
+      return nonce
```
z64:
```suggestion
      nonce
```

PR #207, opened 2020-05-07
URL-Encode request path

MineBartekSA:
The request path should be URL-Encoded.
This change allows writing code like:
```
client.create_reaction(<channel id>, <message id>, "⏭")
```
Now nobody has to keep track of emojis and such.

### DIFF
--- a/src/discordcr/rest.cr
+++ b/src/discordcr/rest.cr
@@ -25,7 +25,6 @@ module Discord
       headers["Authorization"] = @token
       headers["User-Agent"] = USER_AGENT
       headers["X-RateLimit-Precision"] = "millisecond"
-      path = URI.encode(path)
 
       request_done = false
       rate_limit_key = {route_key: route_key, major_parameter: major_parameter.try(&.to_u64)}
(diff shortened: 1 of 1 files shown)


src/discordcr/rest.cr
```diff
@@ -25,6 +25,7 @@ module Discord
       headers["Authorization"] = @token
       headers["User-Agent"] = USER_AGENT
       headers["X-RateLimit-Precision"] = "millisecond"
+      path = URI.encode path
```
greenbigfrog:
```suggestion
      path = URI.encode(path)
```

greenbigfrog (changes requested):
lgtm otherwise :+1:

z64:
I'm not sure this should happen at our base `request` method. It will be redundant in 99% of cases - emoji are the exception where we need to assert that the content is encoded correctly - so I think it might be better to localize this to the relevant REST methods to optimize that path in the simplest way.

MineBartekSA:
I've only added `URI.encode` to the relevant request paths.

PR #208, opened 2020-05-10
remove slash in log to exlude double slash from log output

jackharrhy:
Either `@host : String, @path : String` already contains a `/`, therefore the extra one in the log doesn't seem to be needed :)

image

before ^ (sad)

image

after ^ (happy!)


z64 (approved):
Thanks!

greenbigfrog (approved):

